PR #1, opened 2017-11-01
adding a note about prior gtest in system affecting build

bragadeesh:
Added a note in readme to warn users about existing libgtest.a in the system causing build failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ ctest --output-on-failure
 [sudo] make install
 ```
 
+Note: Existing gtest library in the system (especially static gtest libraries built with other compilers) 
+may conflict with the build process; if errors are encountered with already existing gtest library,
+uninstalling or temporarily renaming gtest will force cmake to download and build gtest and solve the problem
+
 Note: To disable inline assembly optimisations in rocRAND (for both the host library and
 the device functions provided in `rocrand_kernel.h`) set cmake option `ENABLE_INLINE_ASM`
 to `OFF`.


jszuppe:
This happens when `DEPENDENCIES_FORCE_DOWNLOAD` is on, doesn't it?

bragadeesh:
that flag is not being explicitly set; if setting that flag alleviates the problem, I can modify the note accordingly

jszuppe:
If that option is not on, cmake script should first try to find and use installed GTest library and it doesn't download GTest if it finds one.

README.md
```diff
@@ -71,6 +71,10 @@ ctest --output-on-failure
 [sudo] make install
 ```
 
+Note: Existing gtest library in the system (especially static gtest libraries built with other compilers) 
+may conflict with the build process; if errors are encountered with already existing gtest library,
+uninstalling or temporarily renaming gtest will force cmake to download and build gtest and solve the problem
```
jszuppe: You're missing a dot at the end of the sentence.

bragadeesh:
Right, but gtest such a common lib with devs, this is commonly encountered issue,
and this is extra documentation to help developers

I am going to close this PR; I will issue a new PR with updates to contributing guidelines

jszuppe:
Aaa, ok. I thought you've encountered that issue and that's why you're adding that note.

bragadeesh:
Yes, I did encounter that problem, I did not use the flag (which is the cleaner option) to get around the problem. I have updated the note.

PR #3, opened 2017-11-08
Develop 164 tests

arsenm:
I think this should be included in the main set of tests to avoid future regressions

### DIFF
--- a/test/test_xorwow_bug.cpp
+++ b/test/test_xorwow_bug.cpp
@@ -31,6 +31,14 @@
 #define HIP_CHECK(x) ASSERT_EQ(x, hipSuccess)
 #define HIPRAND_CHECK(state) ASSERT_EQ(state, HIPRAND_STATUS_SUCCESS)
 
+#ifdef __AMDGCN__
+#define READ_EXEC() __builtin_amdgcn_read_exec()
+#else
+#define READ_EXEC() -1
+#endif
+
+typedef uint64_t exec_mask_t;
+
 __global__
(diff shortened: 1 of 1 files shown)


jszuppe:
I think we should remove `std::cout`s from the test.

jszuppe:
Also isn't it more test for hcc/ROCm and not for rocRAND? Maybe we shouldn't put tests for compiler into RNG library?

Your commit looks good and doesn't require a discussion, so you can make a separate pull request for that one if you want to have that one in `master` already.

bragadeesh:
@arsenm can you issue the PR to go into develop_164_tests branch; let's discuss separately how to incorporate this testing in hcc  
@jszuppe what @arsenm is asking to do is to not use inline asm for reading the exec register; instead use the built-in intrinsic;

jszuppe:
> @arsenm can you issue the PR to go into develop_164_tests branch; let's discuss separately how to incorporate this testing in hcc

@arsenm already pushed his commits into `develop_164_tests`. This pull request is not from his private fork, but from `develop_164_tests` to `master`.

> @jszuppe what @arsenm is asking to do is to not use inline asm for reading the exec register; instead use the built-in intrinsic;

Sure, we will do that in future. I guess that was the fastest way my colleague knew how to extract values of the exec register at that moment. Also, that test was just to provide minimal test case. By the way, if you want to put that into `hcc` test suite the only dependency on rocRAND is `rocrand_xorwow_precomputed.h`.  I suppose it can be replace by a matrix right in the file with the test (something like `d_xorwow_sequence_jump_matrices[10][800] = { 1234567 };`)

PR #7, opened 2017-12-11
Fix generate_func to pass-by-value

aaronenyeshi:
Seems that generate_func lambda function is being passed into the kernel function by reference. The signature is wrong and this is out of the spec, so generate_func should be passed by value. This is CUDA compliant.


bragadeesh:

PR #8, opened 2017-12-11
Fix m_direction_vectors cast to const uint ptr

aaronenyeshi:
There is a mismatch between this generate_kernel callsite and the function. Cannot take a non-const unsigned int when the function requires a const uint ptr as 3rd argument. generate_kernel = void ( * )(T*, const size_t, **const unsigned int***, const unsigned int, Distribution)

### DIFF
--- a/library/src/rng/sobol32.hpp
+++ b/library/src/rng/sobol32.hpp
@@ -160,7 +160,7 @@ public:
             HIP_KERNEL_NAME(rocrand_host::detail::generate_kernel),
             dim3(blocks_x, blocks_y), dim3(threads), 0, m_stream,
             data, size,
-            m_direction_vectors, m_current_offset,
+            static_cast<const unsigned int*>(m_direction_vectors), m_current_offset,
             distribution
         );
         // Check kernel status


bragadeesh:

jszuppe:
>There is a mismatch between this generate_kernel callsite and the function. Cannot take a non-const unsigned int when the function requires a const uint ptr as 3rd argument. generate_kernel = void ( * )(T*, const size_t, const unsigned int*, const unsigned int, Distribution)

Does it cause any compilation problems? Why is there such a requirement that pass types must exactly match kernel signature? I understand that you can overload function depending on that (pointer to `const T` and pointer to `T` are different types), but why compiler has problem with matching function when there is no overload?

bragadeesh:
the HIP dispatch mechanism and the way hipLaunchKernelGGL is implemented has changed under the covers and it is a bit more strict about types. This HIP change
https://github.com/ROCm-Developer-Tools/HIP/pull/246
has not gone into a public release rocm yet, but it will at some point, so we need to test

aaronenyeshi:
@jszuppe This PR was to fix a compilation problem seen with the latest HIP built from source. We've changed the kernel signature in this PR: https://github.com/ROCm-Developer-Tools/HIP/pull/246/files, please see the tests.

jszuppe:
Ah, so you added a limitation to the kernel launch. Make sure to add a clear note somewhere about this as this behaviour won't be expected by C++ developers.

PR #9, opened 2017-12-11
Fix generate_kernel to pass-by-value

aaronenyeshi:
We should pass GenerateFunc lambda function by value instead of by reference. Cannot pass into kernel function by reference.


bragadeesh:

PR #12, opened 2018-01-17
Workaround to allow TensorFlow ROCm port be built

whchung:


bragadeesh:
@jszuppe do you see any problems with this?
I am going to merge this in; i have tested the builds and there's no problem

jszuppe:
`roc::rocrand` target includes path `/opt/rocm/rocrand/include`, and then `<rocrand.h>` is correct. I assume that in your case you've probably just added include path `/opt/rocm/include`.

IMHO, if we want to change that to `<rocrand/rocrand.h>`, we need to be consistent and change installation path to `<install_prefix>/rocrand/include/rocrand`. Then, `rocrand.h` will be in the same place in both situations: `-I/opt/rocm/include` and `-I/opt/rocm/rocrand/include`. Same thing would have to  happen to `hiprand.h`. I guess, that would be a breaking change for anyone that already is using rocRAND/hipRAND libraries. Alternatively, just include `/opt/rocm/include/rocrand` and `/opt/rocm/include/hiprand` (or `/opt/rocm/rocrand/include` and `/opt/rocm/hiprand/include`).

whchung:
The application we are talking about can NOT include ANY header files under `/opt/rocm` due to its design. All headers are pulled into the build cache at build-time. Right now there seems to be no easy way to change the build system of the application. This PR is the only way to get around it.

I'm fully aware this is not ideal but considering the limited adoption rate of rocRAND/hipRAND I'd like to advocate getting this PR accepted, and have it land on a feature branch (ex: `test_roc_master`), instead of the main branch.

VincentSC:
This MR forces the code to move for _everybody_ to i.e. `/opt/rocm/rocrand/include/rocrand`. I'm sure all kinds of other libraries want it in different places, and with makefiles that should not be a problem.

So when building TensorFlow, can you add `-I ./include/rocrand` in the makefile?

whchung:
@VincentSC no it's not possible at least in the short-term. the build system is much more complicated than adding `-I`. In fact it's not possible to specify any path such as `/somewhere` or `./somewhere`.

I'd recommend accepting this PR and have it merged on a non-master branch. I'll continue to investigate the build structure of the application.

jszuppe:
> The application we are talking about can NOT include ANY header files under /opt/rocm due to its design. All headers are pulled into the build cache at build-time. Right now there seems to be no easy way to change the build system of the application. This PR is the only way to get around it.

What I was just saying is that nowhere in rocRAND there is file under path `rocrand/rocrand.hpp`, only after installing package there is `<install_prefix>/rocrand/rocrand.hpp` path. That's why I am confused about this change.

With that change `hiprand.so` fails to build with `fatal error: 'rocrand/rocrand.h' file not found`.

whchung:
@jszuppe In the application I'm working on, I use pre-built hipRAND and rocRAND on the system, so I have `<install_prefix>/rocrand/include/rocrand.h`, `<install_prefix>/hiprand/include/hiprand_hcc.h`. I haven't tried to build hipRAND / rocRAND.

jszuppe:
> I use pre-built hipRAND and rocRAND on the system,

If you installed it with `apt-get install rocrand`, you should have `<install_prefix>/rocrand/include/rocrand` and `<install_prefix>/include/rocrand` (and similar paths for hiprand).

> I haven't tried to build hipRAND / rocRAND.

I understand that, but how can we recommend merging a change that breaks the build?

VincentSC:
@whchung Ok, I think all is said and clear

I just created a **temporary** branch `tensorflow`. The requested change is in https://github.com/ROCmSoftwarePlatform/rocRAND/blob/tensorflow/library/include/hiprand_hcc.h
There are many more `#include <rocrand.h>` in other files - I assume you're aware of that.

Shall we get back in touch in February, to see if the branch can be removed?

whchung:
@VincentSC / @jszuppe thanks for your help. This solves my immediate issue getting things going. We can regroup a couple of weeks from now. I'm completely not fond of this PR either.

VincentSC:
No worries - your second suggestion to get it solved was simple and clear. 

I looked at TF's build-system and wish you a lot of luck!

bragadeesh:
i think we are all in the same page, I had advised @whchung that this is an inconsistent change; but I was ok with it considering the fact that the /opt/rocm folder structure and users only including /opt/rocm/include in their include directories, i will write an email to discuss this more
anyway @whchung is unblocked for now, thanks

PR #20, opened 2018-06-26
Fix missing gridlaunch.h build issue

aaronenyeshi:
When building with top of 1.8.x branches of HIP and HCC, there is a build issues with missing gridlaunch.h header file. hip::hip_hcc no longer has a dependency on HCC runtime, so we need to use hip::device now.


aaronenyeshi:
@pfultz2

bragadeesh:
@aaronenyeshi 
could you issue PR against develop branch? thanks

aaronenyeshi:
@bragadeesh Thanks, just changed against develop branch now.

PR #27, opened 2018-08-09
Enable using hipcc as the CXX compiler

aaronenyeshi:
If we're using HIP-Clang to compile, we can allow the hipcc as CXX compiler.


jszuppe:
Did you try building rocRAND using hipcc on ROCm platform?

aaronenyeshi:
Yes, it builds with CXX=/opt/rocm/hip/bin/hipcc option added to cmake command. It builds fine, but running into a 19 of 26 test failures on HIP-Clang. We are working on a fix, and will make a PR for that

jszuppe:
> Yes, it builds with CXX=/opt/rocm/hip/bin/hipcc option added to cmake command.

ok, in that case: LGTM, I'll merge in 10 minutes.

> PR for that

This PR will be for rocRAND or for HIP-Clang?

aaronenyeshi:
Still investigating root cause, but more likely for HIP-Clang

PR #29, opened 2018-08-15
Enable hip-clang with cmake changes to link -lpthread and -lm

aaronenyeshi:
This will help fix build issues for tests using hip-clang.


aaronenyeshi:
Fixed with HIP PR https://github.com/ROCm-Developer-Tools/HIP/pull/637

ax3l:
The linked PR is not a fix but a work-around.
If one builds against rocRAND with `clang -x hip` directly, e.g., with recent CMake builds - then rocRAND should declare its public link dependencies directly :)

PR #31, opened 2018-08-23
Make NVGPU_TARGETS a cache variable

### LINKED ISSUE #30: Build issues (module path, CUDA 9 and sm_20)
- `CMAKE_MODULE_PATH=/path/to/hip/` did not work
  - tried various path combinations with `CMAKE_PREFIX_PATH` and `CMAKE_MODULE_PATH` and `HIP_DIR` as suggested by cmake output... however, only a direct `cmake -DHIP_PATH=/path/to/hip/ ..` is working
- when building with CUDA9, lowest CC with `sm_20` yields an error (CUDA 9 no more supports Fermi)
  - can be fixed in `cmake/SetupNVCC.cmake`
  - would be nice, if there would be a cmake option to propagate `NVGPU_TARGETS`
```
Building NVCC (Device) object library/CMakeFiles/hiprand.dir/src/hiprand/hiprand_generated_hiprand_nvcc.cpp.o
nvcc fatal   : Value 'sm_20' is not defined for option 'gpu-architecture'
```
after fixing this, build succeeded.
Test run successfully on P100 and GV100.

Tools: gcc/5.3.0, CUDA9.1, HIP 1.5.18315

jszuppe:
Closes #30


jszuppe:
@tdd11235813 That should work.

tdd11235813:
Thanks, it fixes the issue with `NVGPU_TARGETS` when it is filled with, e.g., `--gpu-architecture=sm_50`.
A simplification could be, that it just takes the compute capabilities, also as list, e.g. "30;35;50".
In cmake such a processing could look like (untested):
```cmake
foreach(CUDA_ARCH_ELEM ${NVGPU_TARGETS})
    list(APPEND HIP_NVCC_FLAGS "--generate-code arch=compute_${CUDA_ARCH_ELEM},code=sm_${CUDA_ARCH_ELEM} --generate-code arch=compute_${CUDA_ARCH_ELEM},code=compute_${CUDA_ARCH_ELEM}")
endforeach()                                                                                                                                                     
```

jszuppe:
Good suggestion. I updated the pull request, it's not as sophisticated as for example [this](https://github.com/BVLC/caffe/blob/master/cmake/Cuda.cmake#L57) code, but it should do the trick.

PR #33, opened 2018-09-05
Fixed performance drop seen with Philox engine with HIP PR 535

### LINKED ISSUE #32: Investigating performance issues in philox
copying @gargrahul 

@jszuppe could you or someone help here? One of the teams is investigating some performance regression issues in rocRAND with changes in compiler (hcc/hip in ROCm). When running (fox example): benchmark_rocrand_kernel --engine philox --dis uniform-float
we are seeing slowdowns. Could you point to the source kernels that are launched with this command? Is it somewhere here:
https://github.com/ROCmSoftwarePlatform/rocRAND/blob/master/library/src/rng/philox4x32_10.hpp
Also explain some top level details about the kernel, and the kernel launch parameters?

It would also help if you could give some tips/ways to simplify the kernel as we track down the perf issues to compiler changes. Thanks

gargrahul:
Fixed performance drop seen with Philox engine with HIP PR  https://github.com/ROCm-Developer-Tools/HIP/pull/535 - Feature native vector types 

Closes #32

### DIFF
--- a/library/include/rocrand_philox4x32_10.h
+++ b/library/include/rocrand_philox4x32_10.h
@@ -195,7 +195,22 @@ public:
     FQUALIFIERS
     unsigned int next()
     {
-        unsigned int ret = (&m_state.result.x)[m_state.substate];
+        unsigned int ret;
+        switch(m_state.substate)
+        {
+            case 0:
+                ret = m_state.result.x;
+                break;
+            case 1:
+                ret = m_state.result.y;
+                break;
+            case 2:
+                ret = m_state.result.z;
+                break;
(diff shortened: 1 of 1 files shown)


bragadeesh:
@ex-rzr , yes please merge to develop and then to master

pramenku:
Can you please merge the fix in master also?

gargrahul:
@VincentSC The fix is already part of master branch https://github.com/ROCmSoftwarePlatform/rocRAND/blob/master/library/include/rocrand_philox4x32_10.h#L199

PR #48, opened 2019-02-13
Merging Develop to Master for rocm-2.1 

saadrahim:
Need to branch for QA testing of rocm-2.1. Are there any concerns with this? I can base of the current master but it may be worth while to include the changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,6 @@ local.properties
 # TeXlipse plugin
 .texlipse
 CMakeLists.txt.user
+
+### Gtilab CI ###
+.gitlab-ci-gputest.yml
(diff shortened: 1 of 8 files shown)


ex-rzr:
Why was the mode set to "Squash and merge" instead of "Create a merge commit"?

ex-rzr:
Weird, GitHub sets it to "Squash and merge" again when I reload the page. Any ideas why?

saadrahim:
> Why was the mode set to "Squash and merge" instead of "Create a merge commit"?

I was just checking what merge options were available. I didn't want to make another squash mistake. Weird that it even saved when I pulled the drop down. I don't intent to repeat a the mistake on rocPRIM>

ex-rzr:
Yeah, very confusing :/ I will consider this strange GitHub's behavior in the future.

Issue #39, opened 2018-11-28
rocrand.h file not found


ntrost57:
`library/include/hiprand_hcc.h` should include `rocrand/rocrand.h` instead of `rocrand.h`.

ex-rzr:
Fixed in #58

Issue #38, opened 2018-11-27
rocrand-config.cmake directory


ntrost57:
rocrand is (as far as I know) the only library that installs its `rocrand-config.cmake` and related files in `<rocm-dir>/lib/cmake/rocrand/rocrand/`
All other libraries install to `<rocm-dir>/lib/cmake/<roclib>/`
Is there a special reason for that?

ex-rzr:
This is very strange.
How do you build and install rocRAND?

I've just checked three ways:
1. `sudo make install`
2. `make package`, then `sudo dpkg -i rocrand-1.8.2-Linux.deb `
3. `sudo apt install rocrand`

All commands create `/opt/rocm/rocrand/lib/cmake/rocrand/rocrand-config.cmake` and `/opt/rocm/rocrand/include/rocrand.h`, i.e. without this strange `rocrand/rocrand` path.

ntrost57:
It is about `/opt/rocm/lib/cmake/` where all symlinks are created. Here, the directory is wrong.

ex-rzr:
Fixed in #58

PR #51, opened 2019-03-14
Fix the rocrand header path

zjunweihit:
After install the debian package, rocrand header files
locate rocrand folders rather than the root of include.

### DIFF
--- a/library/include/hiprand_hcc.h
+++ b/library/include/hiprand_hcc.h
@@ -21,7 +21,7 @@
 #ifndef HIPRAND_HCC_H_
 #define HIPRAND_HCC_H_
 
-#include <rocrand.h>
+#include <rocrand/rocrand.h>
 
 typedef rocrand_generator_base_type hiprandGenerator_st;
 
--- a/library/include/hiprand_kernel_hcc.h
+++ b/library/include/hiprand_kernel_hcc.h
@@ -34,7 +34,7 @@
 /// \cond
 #define FQUALIFIERS QUALIFIERS
 /// \endcond
-#include <rocrand_kernel.h>
+#include <rocrand/rocrand_kernel.h>
 
 /// \cond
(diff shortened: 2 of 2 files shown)


zjunweihit:
Error log:
```
In file included from /opt/rocm/include/hiprand/hiprand.h:55:
/opt/rocm/include/hiprand/hiprand_hcc.h:24:10: fatal error: 'rocrand.h' file not found
#include <rocrand.h>

...
file included from /opt/rocm/include/hiprand/hiprand_kernel.h:58:
/opt/rocm/include/hiprand/hiprand_kernel_hcc.h:37:10: fatal error: 'rocrand_kernel.h' file not found
#include <rocrand_kernel.h>
```

ex-rzr:
These changes break building libhiprand.so if the deb is not installed.

I think this issue may need changes in https://github.com/ROCmSoftwarePlatform/rocRAND/blob/master/cmake/Common.cmake#L15
or perhaps even replacing it with https://github.com/RadeonOpenCompute/rocm-cmake if possible.

#38 and #39 are related to this issue too.

saadrahim:
> These changes break building libhiprand.so if the deb is not installed.
> I think this issue may need changes in https://github.com/ROCmSoftwarePlatform/rocRAND/blob/master/cmake/Common.cmake#L15
> or perhaps even replacing it with https://github.com/RadeonOpenCompute/rocm-cmake if possible.
> #38 and #39 are related to this issue too.

Hi @VincentSC ,

What is that status of this issue? Will replacing with https://github.com/RadeonOpenCompute/rocm-cmake solve this? Can we aim to do this before May 8th? Otherwise it will have to be pushed into the next release?

Saad

saadrahim:
@VincentSC  what is the best way to solve this?

VincentSC:
This needs to be retested with rocm-cmake in place. Since that task has been finished, can the retest be done?

ex-rzr:
See #58

PR #60, opened 2019-06-07
Standardizing include path to include the library name

saadrahim:
rocrand used to install to /opt/rocm/include/rocrand and /opt/rocm/include/hiprand
After switching to rocm-cmake, rocrand would place all the headers in /opt/rocm/include without the directories.
These changes restore that behavior with rocm-cmake.

PR #58, opened 2019-05-31
 Use rocm-cmake

ex-rzr:
Fixes #38, fixes #39, fixes #44

PR #63, opened 2019-06-14
Revert "Standardize include path to include the library name (#60)"

yoichiyoshida:
Reverts ROCmSoftwarePlatform/rocRAND#62
This change breaks upstream projects PyTorch and Caffe2. Reverting breaking change until a fix is available.


ex-rzr:
Maybe it's better to reset the branch to the previous commit? There are no branches based on the current master: https://github.com/ROCmSoftwarePlatform/rocRAND/network

saadrahim:
Did we go back far enough that Tensorflow will not have a problem? Starting to use rocm-cmake started this?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

yoichiyoshida:
I can reset the branch back to 2.5, which is before rocm-cmake. There is one branch in between by Akila, but hopefully that is minor enough that it can be recreated.

VincentSC:
Please hold any action untill it is fully clear what's the problem. The reason for #60 was to standardize the location for the ROCm libraries. If now something breaks, then probably that has to be fixed, not?

Can you explain why you want to revert? What is going wrong exactly? What are we missing?

yoichiyoshida:
The new standardized location is different than the location for most of the other projects at the moment. It's too close to release to ask upstream projects to arbitrarily change their paths without communication. #60/62 breaks PyTorch (but fixes Tensorflow), and #58/59 breaks Tensorflow, so the plan is to roll it back to before those two so both upstream projects stay working for this release (ie. back to 2.5). After the 2.6 release these changes can be revisited so that both upstream projects can be made to work.

PR #66, opened 2019-07-03
Added hip-clang jobs

amdkila:
- Adds to CI testing
- Needs to pass CI


akilaMD:
There are 4 test failures (out of 26) when using the hip-clang compiler. Otherwise, this passes CI. I would suggest that we merge this.

PR #69, opened 2019-07-15
Rocmcmake

AlexBrownAMD:
Fix install paths and symlinks with rocm-cmake.


ex-rzr:
Could you elaborate what is AMD's final decision on the library's paths?
We do not know the current goal and why exaclty the previous solution didn't work (except that some libraries needed some work to use new paths).
Do I understand it correctly that we keep rocm-cmake (#58), but revert new paths (#60), i.e. use apply again #62? So it's rocm-cmake+old paths, correct?

saadrahim:
> Could you elaborate what is AMD's final decision on the library's paths?
> We do not know the current goal and why exaclty the previous solution didn't work (except that some libraries needed some work to use new paths).
> Do I understand it correctly that we keep rocm-cmake (#58), but revert new paths (#60), i.e. use apply again #62? So it's rocm-cmake+old paths, correct?

Yes, that is the plan. Old paths must be maintained.

saadrahim:
This approach may break libraries that consume rocrand. We will wait to switch to rocm-cmake.

PR #72, opened 2019-07-19
Fix hip-clang compilation error.

mhbliao:
- Need to include <functinoal> explicitly.
- Need to replace `__attribute__((cpu))` with `__host__`.

### DIFF
--- a/test/crush/stat_test_common.hpp
+++ b/test/crush/stat_test_common.hpp
@@ -29,6 +29,7 @@
 #include <numeric>
 #include <utility>
 #include <algorithm>
+#include <functional>
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
Thanks @amdkila , can this be cherry-picked into `master` branch or `master-rocm-2.6` branch?

mhbliao:
could you please merge this into 'master' branch as well. it's urgent as it currently block the centos ci.

saadrahim:
That can be done by end of day. We have some other updates we are planning to add. Is that reasonable?

mhbliao:
ok, that sounds good.

PR #85, opened 2019-09-03
Adding sles support

saadrahim:


Jenkinsfile
```diff
@@ -11,7 +11,7 @@ rocRANDCI:
 
     def rocrand = new rocProject('rocRAND')
 
-    def nodes = new dockerNodes(['gfx900 && ubuntu', 'gfx906 && ubuntu', 'gfx900 && centos7', 'gfx906 && centos7'], rocrand)
+    def nodes = new dockerNodes(['gfx900 && ubuntu', 'gfx906 && ubuntu', 'gfx900 && centos7', 'gfx906 && centos7', 'gfx900 && sles'], rocrand)
```
amdkila: wasn't the plan for sles label to be 'sles' instead of 'gfx900 && sles'?

docker/dockerfile-install-sles
```diff
@@ -0,0 +1,21 @@
...
+
+FROM ${base_image}
+LABEL maintainer="saad.rahim@amd"
+
+# Copy the rpm package of rocfft into the container from host
+COPY *.rpm /tmp/
+
+# Install the rpm package, and print out contents of expected changed locations
+RUN zypper -n update && zypper install -n\
+    /tmp/rocfft-*.rpm \
```
amdkila: I'm not sure what the docker install file does, but I know that rocJenkins looks for it. Should change all the instances of rocfft to rocrand

PR #94, opened 2019-10-11
Switching to hipcc as the compiler

saadrahim:


install
```diff
@@ -105,10 +105,7 @@ fi
 # before 'cmake' or setting cmake option 'CMAKE_CXX_COMPILER' to path to the HCC compiler.
```
feizheng10: change the comments as well

install
```diff
@@ -105,10 +105,7 @@ fi
...
 #
 
-compiler="hcc"
-if ($build_hip_clang); then
-    compiler="hipcc"
-fi
+compiler="hipcc"
```
feizheng10: how could we tell it is hip-clang?

saadrahim: not sure

PR #99, opened 2019-11-14
rocRAND changes to support multiple ROCM installation

pruthvistony:
- New mode of building is added "-r,--relocatable" which is used
      for ROCm stack installed in /opt/rocm-ver.
- Below CMAKE parameters are set/overwritten in above mode
      CMAKE_INSTALL_PREFIX
      CMAKE_PREFIX_PATH
      CMAKE_SHARED_LINKER_FLAGS
      CMAKE_MODULE_PATH
      ROCM_DISABLE_LDCONFIG

Signed-off-by: Pruthvi Madugundu <mpruthvi@gmail.com>

### DIFF
--- a/cmake/Common.cmake
+++ b/cmake/Common.cmake
@@ -1,4 +1,6 @@
 #
+set(ROCM_DISABLE_LDCONFIG OFF CACHE BOOL "")
+
 function(package_set_postinst_prerm LIB_NAMES LIB_DIRS INCLUDE_DIRS)
     list(LENGTH LIB_NAMES len1)
     list(LENGTH LIB_DIRS  len2)
@@ -18,7 +20,9 @@ function(package_set_postinst_prerm LIB_NAMES LIB_DIRS INCLUDE_DIRS)
 
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nmkdir -p ${inc_dir}/../../include/")
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nmkdir -p ${lib_dir}/../../lib/cmake/${lib_name}")
-        set(POSTINST_SOURCE "${POSTINST_SOURCE}\necho \"${lib_dir}\" > /etc/ld.so.conf.d/${lib_name}.conf")
+        if(NOT ${ROCM_DISABLE_LDCONFIG})
+            set(POSTINST_SOURCE "${POSTINST_SOURCE}\necho \"${lib_dir}\" > /etc/ld.so.conf.d/${lib_name}.conf")
+        endif()
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nln -sr ${inc_dir} ${inc_dir}/../../include/${lib_name}")
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nln -sr ${lib_dir}/lib${lib_name}.so ${lib_dir}/../../lib/lib${lib_name}.so")
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nln -sr ${lib_dir}/cmake/${lib_name} ${lib_dir}/../../lib/cmake/${lib_name}\n")
--- a/install
+++ b/install
@@ -11,6 +11,7 @@ function display_help()
     echo "    [-h|--help] prints this help message"
     echo "    [-i|--install] install after build"
     echo "    [-p|--package] build package"
+    echo "    [-r]--relocatable] create a package to support relocatable ROCm"
     #Not implemented yet
     #    echo "    [-d|--dependencies] install build dependencies"
     echo "    [-c|--clients] build library clients too (combines with -i & -d)"
@@ -28,7 +29,9 @@ build_clients=false
 build_release=true
 run_tests=false
 build_hip_clang=false
-rocm_path=/opt/rocm/bin
+rocm_path=/opt/rocm
+build_relocatable=false
+
 # #################################################
 # Parameter parsing
 # #################################################
@@ -36,7 +39,7 @@ rocm_path=/opt/rocm/bin
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
-    GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,clients,debug,hip-clang,test,package --options hicdtp -- "$@")
+    GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,clients,debug,hip-clang,test,package,relocatable --options hicdtpr -- "$@")
 else
     echo "Need a new version of getopt"
     exit 1
@@ -62,6 +65,9 @@ while true; do
 	-p|--package)
 	    build_package=true
 	    shift ;;
+        -r|--relocatable)
+            build_relocatable=true
+            shift ;;
 	-c|--clients)
 	    build_clients=true
 	    shift ;;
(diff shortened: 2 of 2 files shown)


cmake/VerifyCompiler.cmake
```diff
@@ -35,11 +35,11 @@ elseif(HIP_PLATFORM STREQUAL "hcc")
     else()
         # Workaround until hcc & hip cmake modules fixes symlink logic in their config files.
         # (Thanks to rocBLAS devs for finding workaround for this problem.)
-        list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hcc /opt/rocm/hip)
-        # Ignore hcc warning: argument unused during compilation: '-isystem /opt/rocm/hip/include'
+        list(APPEND CMAKE_PREFIX_PATH ${CMAKE_INSTALL_PREFIX}/hcc ${CMAKE_INSTALL_PREFIX}/hip)
```
srinivamd: Same comment as for rocPRIM:
Does this impact the build and install documentation for this component?

What happens when this component package is built without CMAKE_INSTALL_PREFIX and installed in a multi ROCm installed system?

install
```diff
@@ -28,7 +28,12 @@ build_clients=false
 build_release=true
 run_tests=false
 build_hip_clang=false
-rocm_path=/opt/rocm/bin
+
+rocm_path=/opt/rocm
+if ! [ -z ${ROCM_INSTALL_PATH+x} ]; then
+    rocm_path=${ROCM_INSTALL_PATH}
```
srinivamd: How about adding a command-line option, say, --prefix, to specify where to install instead of relying on an undocumented environment variable?

library/CMakeLists.txt
```diff
@@ -273,6 +279,7 @@ package_set_postinst_prerm(
     "hiprand;rocrand"
     "${CMAKE_INSTALL_PREFIX}/hiprand/lib;${CMAKE_INSTALL_PREFIX}/rocrand/lib"
     "${CMAKE_INSTALL_PREFIX}/hiprand/include;${CMAKE_INSTALL_PREFIX}/rocrand/include"
+    "/opt/rocm/hiprand/lib;/opt/rocm/rocrand/lib"
```
srinivamd: Is /opt/rocm ok to use here?

test/package/CMakeLists.txt
```diff
@@ -30,9 +30,9 @@ if(HIP_PLATFORM STREQUAL "nvcc")
 endif()
 
 # Find rocRAND
-find_package(rocrand REQUIRED CONFIG HINTS ${rocrand_DIR} PATHS "/opt/rocm/rocrand")
+find_package(rocrand REQUIRED CONFIG HINTS ${rocrand_DIR} PATHS "${CMAKE_INSTALL_PREFIX}/rocrand")
 # Find hipRAND
-find_package(hiprand REQUIRED CONFIG HINTS ${hiprand_DIR} PATHS "/opt/rocm/hiprand")
+find_package(hiprand REQUIRED CONFIG HINTS ${hiprand_DIR} PATHS "${CMAKE_INSTALL_PREFIX}/hiprand")
```
srinivamd: Not sure what "changes to support reloctable ROCM installation" means. An explanation of what that means would be helpful to evaluate these changes.
Can CMAKE_INSTALL_PREFIX could be set to "/opt/rocm" where /opt/rocm is a symlink? 
If prefix is set to /opt/rocm will be package overwrite the symlink? Probably should test those scenarios, and explain how "relocatable" installation is supported in documentation.

cmake/VerifyCompiler.cmake
pruthvistony: CMAKE_INSTALL_PREFIX is default set to /opt/rocm, so if a user want to build the rocRAND to a new path, then CMAKE_INSTALL_PREFIX needs to be set.
Regarding building when multiple ROCm version are installed, need to set /opt/rocm link to the required version and CMAKE_INSTALL_PREFIX can be set/unset, accordingly the rocRAND can be built.

library/CMakeLists.txt
pruthvistony: Yes. This parameter is used to write the ld.conf.so entry, which is preferred to have it as /opt/rocm. So any external application/framework want to use this lib, will be able to find it through ldconf entry with the symlink to the required version of rocm.

test/package/CMakeLists.txt
pruthvistony: Its actually multiple ROCM installation, I have updated the comment.
When CMAKE_INSTALL_PREFIX is set /opt/rocm and it is symlink.  The symlink is not overwritten, the packages gets installed into the folder it is pointing too.

srinivamd: Would it be ok to overwrite the contents of /opt/rocm symlink? Could that potentially overwrite the incorrect rocm release with the package contents, say, when a built package is installed weeks later by when /opt/rocm symlink could have been modified? Could make it difficult to track install errors I think.

library/CMakeLists.txt
saadrahim: I am not sure about this. It seems we are hard coding it. Also, this will not work with SPACK. We should no assume that anything exists in /opt/rocm

saadrahim:
Please change the target to develop for this PR

library/CMakeLists.txt
saadrahim: This does not make a package relocatable. We should not hard code this to /opt/rocm/

cmake/VerifyCompiler.cmake
saadrahim: CMAKE_INSTALL_PREFIX should not be used to search for tools required to build the library. This shold be a separate variable.

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended
 list(APPEND CMAKE_MODULE_PATH
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules
-    ${HIP_PATH}/cmake /opt/rocm/hip/cmake # FindHIP.cmake
+    ${HIP_PATH}/cmake ${CMAKE_INSTALL_PREFIX}/hip/cmake # FindHIP.cmake
```
saadrahim: This should not be CMAKE_INSTALL_PREFIX. We can install the library to separate location other than the rocm location.

cmake/Common.cmake
```diff
@@ -15,10 +15,11 @@ function(package_set_postinst_prerm LIB_NAMES LIB_DIRS INCLUDE_DIRS)
         list(GET LIB_NAMES ${val} lib_name)
         list(GET LIB_DIRS  ${val} lib_dir)
         list(GET INCLUDE_DIRS ${val} inc_dir)
+        list(GET LDCONF_ENTRYS ${val} ldconf_entry)
 
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nmkdir -p ${inc_dir}/../../include/")
         set(POSTINST_SOURCE "${POSTINST_SOURCE}\nmkdir -p ${lib_dir}/../../lib/cmake/${lib_name}")
-        set(POSTINST_SOURCE "${POSTINST_SOURCE}\necho \"${lib_dir}\" > /etc/ld.so.conf.d/${lib_name}.conf")
+        set(POSTINST_SOURCE "${POSTINST_SOURCE}\necho \"${ldconf_entry}\" > /etc/ld.so.conf.d/${lib_name}.conf")
```
saadrahim: I need time to review this change.

CMakeLists.txt
pruthvistony: Will make the changes as discussed in rocPRIM review and will add ROCM_PATH.

cmake/VerifyCompiler.cmake
pruthvistony: Yes will add ROCM_PATH similar to rocPRIM changes.

library/CMakeLists.txt
pruthvistony: Yes, it is hard coding. As per the current design the loader will search all the dependent libs based on RUNPATH using $ORIGIN. It will not use ldconf entries or LD_LIBRARY_PATH, as per the customer request. The above changes are required for normal users who have /opt/rocm symlink and they will use the ldconf entries to load the required libraries.

pruthvistony: Same comment as above.

cmake/Common.cmake
pruthvistony: The changes are mainly to add a new parameter and this parameters adds the ld.so.conf entry which is passed as /opt/rocm.

pruthvistony:
I have updated the PR to master-rocm-2.10.1 branch and updated the commit with all the review comments.

pruthvistony:
@saadrahim 
Please review the changes.

PR #116, opened 2020-04-01
[WIP] Tweaked CMake files to compile/run benchmarks with clang on gfx1010/1012

mihirparadkar:
I managed to build and run the benchmark kernel on an RX 5700 XT (gfx1010) with latest stable LLVM+Clang (10.0.0), the included tweaks to the CMake files, and editing some of the generated Makefiles manually.

System is Linux Mint 19.3 with linux kernel 5.5.11-050511-generic, amdgpu drivers, Ryzen 3900X CPU and ROCm version 3.1.0 .

CMake command: `CXX=hipcc cmake -DBUILD_BENCHMARK=ON -DENABLE_INLINE_ASM=OFF ../.`

Environment Variables: 
```shell
export HIP_PLATFORM=clang
export ROCM_PATH=/opt/rocm
export HIP_CLANG_PATH=<path-to-llvm-10.0.0 >/bin/
export DEVICE_LIB_PATH=/opt/rocm/lib
```
The generated `flags.make` and `link.txt` files contained invalid flags for clang, specifically the -hc flag.
(flags.make) 
`CXX_FLAGS =  -Wno-unused-command-line-argument -Wall -Wextra -O3 -DNDEBUG -fPIC   -hc -fPIC -std=c++14 -std=c++11`
(link.txt)
[...]

PR #117, opened 2020-04-03
Add support for HIP_PLATFORM=clang

amdkila:
- I haven't tested this yet...please don't merge


saadrahim:
Did you see pr116

saadrahim:
Did you see pr116

amdkila:
I just saw it now. I suspect the user made the one change that was necessary to allow rocRAND to build. However, tests and packaging are a different story

mihirparadkar:
@amdkila that is correct. The changes in #116 are more of a call-for-help since I have minimal experience with cmake (those changes require manual edits to Makefiles after the cmake command to build/run the benchmark and CuPy random array creation). These changes (plus adding gfx1010/1011/1012 to AMDGPU_TARGETS) seem like the correct way to fix my problem.

I'm happy to test this change for Linux Mint 19.3 and RX 5700XT + RX 5500XT.

amdkila:
This ended up passing CI and fixes the clang issues. It should be safe to merge.

library/src/hiprand/hiprand_clang.cpp
```diff
@@ -0,0 +1,382 @@
+// Copyright (c) 2017 Advanced Micro Devices, Inc. All rights reserved.
```
saadrahim, line 1: What is this file for?
I always bug people about copyright dates.

amdkila: Good question. So, there's a part of the code where a file with hiprand_${HIP_PLATFORM} is used to read a bunch of hiprand functions that differ from hip and nvcc. In the library/src/hiprand directory, there are currently 2 files- hiprand_hcc.cpp and hiprand_nvcc.cpp. 

I just copied and pasted the hcc file and renamed it to _clang.cpp, so we can have both covered. I know, not an elegant solution, but it's the only one I figured out. I will fix the copyright dates.

mihirparadkar: Just curious, is there anything HCC-specific in hiprand_hcc.cpp/hiprand_hcc.hpp ? In #116 I just made cmake point to hiprand_hcc.* when `${HIP_PLATFORM}` is `clang`

saadrahim: No there is isn't, so your comment is noted that we shouldn't have two files.

saadrahim: @amdkila don't bother with copyright dates, we should be able to have one file.

saadrahim:
@amdkila Can we close this?

PR #124, opened 2020-05-07
Ctest command change

saadrahim:


.jenkins/common.groovy
```diff
@@ -20,12 +20,14 @@ def runCompileCommand(platform, project, jobName)
...
     String sudo = auxiliary.sudo(platform.jenkinsLabel)
+    String centos = platform.jenkinsLabel.contains('centos') ? '3' : ''
+    def testCommand = "ctest${centos} --output-on-failure"
 
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}/build/release
                 make -j4
-                ${sudo} LD_LIBRARY_PATH=/opt/rocm/lib/ ctest --output-on-failure
+                ${sudo} LD_LIBRARY_PATH=/opt/rocm/lib/ ${testCommand} --output-on-failure
```
eidenyoshida, line 30: Looks like you have --output-on-failure twice

saadrahim: I don't see it? Looks like once to me?

eidenyoshida: Definition of testCommand has it too

PR #134, opened 2020-07-03
Fixing package naming to reflect OS name and architecture

saadrahim:


neon60:
Looks good. Checking the new package.

PR #136, opened 2020-07-27
Removing docker files no longer needed for CI and adding static library to CI

saadrahim:


CMakeLists.txt
```diff
@@ -25,11 +25,15 @@ cmake_minimum_required(VERSION 3.5.1 FATAL_ERROR)
 # Install prefix
 set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories")
 
+#Adding CMAKE_PREFIX_PATH
+list( APPEND CMAKE_PREFIX_PATH /opt/rocm/llvm /opt/rocm )
```
eidenyoshida, line 29: Maybe you also want to add ${ROCM_PATH} ${ROCM_PATH}/llvm for relocatable installs

saadrahim: I don't support that. It should be handles in cmake arguments, not in cmake.

Issue #140, opened 2020-08-25
Invalid use of inline assembly for GFX1010 target


rigtorp:
**Describe the bug**

When building PyTorch the following error is observed:
```
[ 86%] Building HIPCC object caffe2/CMakeFiles/torch_hip.dir/__/aten/src/THHUNN/torch_hip_generated_RReLU.hip.o
<inline asm>:1:24: error: invalid operand for instruction
        v_mad_u64_u32 v[2:3], s[10:11], s1, v42, v[8:9]
                              ^
note: !srcloc = 13700931
<inline asm>:1:24: error: invalid operand for instruction
        v_mad_u64_u32 v[4:5], s[10:11], s46, v42, v[8:9]
```

I traced this down to use of inline assembly in rocRAND:
```
/opt/rocm-3.7.0/rocrand/include/rocrand_common.h
61:    asm volatile("v_mad_u64_u32 %0, %1, %2, %3, %4"
```

If I comment out the above line, the PyTorch build proceeds.

**To Reproduce**
Build PyTorch with ROCm, see ROCmSoftwarePlatform/pytorch#718

**Expected behavior**
Build succeeds.

doctorcolinsmith:
@rigtorp Thanks for reporting the issue.  We're working to fix the invalid use of v_mad_u64_u32 on gfx10 series hardware in an upcoming release.  You may see this same error recur in other ROCm libraries.

saadrahim:
A fix has been merged into the develop branch. @rigtorp Please verify if this resolves your issue. I will assume the issue is resolved in a week and close the ticket if no response is received.

rigtorp:
Yes PyTorch builds for GFX1010 now. But I cannot confirm functionality until https://github.com/ROCmSoftwarePlatform/rocPRIM/issues/180 has been fixed.

saadrahim:
Work on that issue is ongoing.

PR #142, opened 2020-09-01
Add static analysis CI job with format check

amdkila:


saadrahim:
Planning to add clang-format to checks to ensure consistent code formatting. Please take a look and see if there are any objections.

saadrahim:
@amdkila Do we need to add the .clang-format file for customization or are we ok with defaults?

amdkila:
The defaults should be fine I think- the most cotntroversial is the SortIncludes, but that doesn't break functionality for rocRAND (I tested building/packaging/running tests locally). And sure, I'll review the updated githooks when your PR is up

.jenkins/staticanalysis.groovy
```diff
@@ -11,7 +11,7 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     def command = """#!/usr/bin/env bash
             set -ex
             cd ${project.paths.project_build_prefix}
-            doxygen docs/Doxyfile
+            doxygen doc/Doxyfile
```
saadrahim, line 14: I would rename the folder to docs. All other libraries it is called docs. Just some standardization to make life easier.

saadrahim:
Solves #140

saadrahim:
inactivity.

PR #148, opened 2020-10-15
Change xnack to off

eidenyoshida:


doctorcolinsmith:
As a sanity check, we're looking for xnack on and off, which produces kernels.   The any mode produces a single kernel that works in both modes.

saadrahim:
I thought we are just merging in changes to default to xnack off.

saadrahim (approved):

ax3l:
@ eidenyoshida  @saadrahim 

I think this PR caused a regression in our CI, [building rocRAND like this](https://github.com/AMReX-Codes/amrex/blob/31ca23418f488fe0b309ef3fc3e99c522ced4921/.github/workflows/dependencies/dependencies_hip.sh#L27-L52):
```bash
mkdir build
cd build
cmake .. -DCMAKE_CXX_COMPILER=$(which hipcc)
```
and does not build anymore with:
```
[ 60%] Building CXX object library/CMakeFiles/rocrand.dir/src/rocrand.cpp.o
clang-11: error: Unsupported CUDA gpu architecture: gfx900:xnack-
clang-11: error: Unsupported CUDA gpu architecture: gfx906:xnack-
clang-11: error: Unsupported CUDA gpu architecture: gfx908:xnack-
```

doctorcolinsmith:
@ax3l Thanks for reporting the regression.  Until it is fixed, we recommend using the most recent released version instead of the master branch for a stable build.

PR #153, opened 2020-11-17
Updates to the package name to support ROCm upgrade

pruthvistony:
- Git commit number is removed
- ROCM_VERSION number will be appended to component version as ...<rocm_version_number>
  only when defined in the ENV.
- CPACK_DEBIAN_PACKAGE_RELEASE and CPACK_RPM_PACKAGE_RELEASE are set to git hash,
  if not defined in the ENV which will be defined in the CI environment to provide all the build info.

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -287,6 +287,16 @@ endif()
 set(CPACK_DEBIAN_FILE_NAME "DEB-DEFAULT")
 set(CPACK_RPM_FILE_NAME "RPM-DEFAULT")
 
+# Below block of code is to support older version of rocm-cmake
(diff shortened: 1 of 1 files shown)


pruthvistony:
@saadrahim, 
Please review these changes.
For reference - Changes into rocm-cmake are merged - https://github.com/RadeonOpenCompute/rocm-cmake/commit/2d2b479899397263bf60867539492ad581ed6c66

pruthvistony:
The rocm-cmake code is not latest in the dockers which are used for testing, due to which the errors are seen.
I am planning to add error handling code, so that it works with older and newer version of rocm-cmake

saadrahim:
@stanleytsang-amd Can you look at this? I have no objections.

stanleytsang-amd (approved):

PR #154, opened 2020-11-18
Update master to include PR #153

stanleytsang-amd:
This needs to be in mainline ASAP for ROCm 4.0.


saadrahim:
Master is only fast forwarded from staging after tests pass.

PR #155, opened 2020-11-20
Handling the GIT TAG null case for package file name

pruthvistony:
The code changes handle the case when rocRAND_VERSION_TWEAK variable is not defined,
(set by the rocm_setup_version() API), happens when GIT TAG is not available in the repo.


pruthvistony:
@stanleytsang-amd, @doctorcolinsmith and @saadrahim ,
Please review this PR. Thank you.

stanleytsang-amd (approved):

PR #158, opened 2020-12-02
Adding backward compatibility for target-id syntax for AMDGPU_TARGETS

saadrahim:


stanleytsang-amd:
LGTM. Is this something we should implement for all math libs?  Or is this in response to something specific for rocRAND?

saadrahim:
We should do this for all libraries. I prototyped on rocRAND.

saadrahim:
@t-tye  If you confirm that this is a suitable detection method, math libraries can implement on all libraries.  I tested on a recent mainline build and rocm 3.10. Code is working as expected.

saadrahim:
Sorry, one more change is needed to ensure we have correct compiler.

saadrahim:
This is ready for review.

CMakeLists.txt
```diff
@@ -53,8 +53,17 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE CACHE BOOL "Add paths to linker searc
 # Set the default value of BUILD_SHARED_LIBS
 set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared")
 
-# AMD targets
-set(AMDGPU_TARGETS gfx900:xnack-;gfx906:xnack-;gfx908:xnack- CACHE STRING "List of specific machine types for library to target")
+# Detect compiler support for target ID
+set(TARGET_ID_COMMAND bash "-c" "/opt/rocm/bin/hipcc --help | grep mcode-object-version" )
```
dcyca: Is there a way to invoke hipcc without making assumptions about the install location?

TorreZuk (commented):
This might be simpler as not sure of bash and grep on windows (but will probably have them):
if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
  # Determine if hip has feature
  execute_process(COMMAND ${CMAKE_CXX_COMPILER} "--help" OUTPUT_VARIABLE CXX_OUTPUT
                  OUTPUT_STRIP_TRAILING_WHITESPACE
                  ERROR_STRIP_TRAILING_WHITESPACE)
Then use a cmake regex to look for flag, but hope Tony might say you can look for a version rather than flag in help.  The could use --version instead of --help.

CMakeLists.txt
saadrahim: I changed it to the CMAKE_CXX_COMPILER.

TorreZuk (approved):
LGTM.   I'll comment in the channel as going forward I would think we should also have a custom set option based on environment variable override ( I suggested this in the original xnack thread but didn't seem to get anyone to buy in ).

PR #160, opened 2020-12-22
Benchmark performance workaround + install script fixes

stanleytsang-amd:
This workaround fixes the 40% performance drop observed in rocRAND benchmarks.

I've also enabled the -d/--dependency flag in the install script to allow the user to force install GTest, in the case where GTest is already installed in a system folder (which rocRAND picks up) but is too outdated and results in compilation errors.

I also fixed some bugs with debug builds in the install script, it should now properly build in debug mode with -g.  Previously it did nothing different aside from creating the debug build folder.


library/include/rocrand_mtgp32.h
```diff
@@ -128,9 +128,9 @@ typedef mtgp32_params_fast_t mtgp32_fast_params;
 
 struct mtgp32_state
 {
-    unsigned int status[MTGP_STATE];
     int offset;
     int id;
+    unsigned int status[MTGP_STATE];
```
saadrahim, line 133: You are just changing the order. Is this the workaround?

stanleytsang-amd: Yes.  Changing the order of variables seems to affect how the struct is packed.

saadrahim: Rather silly, not sure if you filed a bug.

saadrahim (approved):

library/include/rocrand_mtgp32.h
stanleytsang-amd: Agreed, it seems related to the targetid changes, I'll file an internal ticket later.

TorreZuk: This is interesting.   Did you profile to see what functions were mostly affected by the struct alignment changes?

stanleytsang-amd: The biggest performance changes were seen in the device kernel functions, specifically for the uniform-uint, uniform-double, and uniform-float types (close to 40%).  Normal-float and log-normal float had smaller performance hits (< 10%).

PR #161, opened 2021-01-12
Address Sanitizer

eidenyoshida:


saadrahim (commented):
Looks good, need to see results on CI

eidenyoshida:
Debug build is supposed to run with address sanitizer enabled. Do we expect to see extra verbose output from our tests when this is enabled?

saadrahim:
Only if it detects something. Not sure if there is a default output.

CMakeLists.txt
```diff
@@ -53,6 +53,20 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE CACHE BOOL "Add paths to linker searc
...
 set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared")
 
+# Set the default value of BUILD_SHARED_LIBS
+set(ADDRESS_SANITIZER OFF CACHE BOOL "Build with address sanitizer")
+
+if ( ${ADDRESS_SANITIZER} )
+    set ( HSAKMT_C_FLAGS "${HSAKMT_C_FLAGS}" -fsanitize=address )
+    set ( HSAKMT_LINK_FLAGS "${HSAKMT_LINK_FLAGS}" -fsanitize=address )
+    if ( BUILD_SHARED_LIBS} )
+        set ( HSAKMT_LINK_FLAGS "${HSAKMT_LINK_FLAGS}" -shared-libasan )
```
saadrahim, line 63: we should change the variable name. HSAKMT is a rocm project. Let's call it rocRAND or something generic.

saadrahim (changes requested):

CMakeLists.txt
saadrahim: Code needs to move down. we need to customize for us.

b-sumner:
I think this looks fine (but agree that HSAKMT can be changed).  I am wondering though about a debug build without the sanitizer.  The sanitizer runtime can add a fair amount of memory overhead (in addition to the other runtime overhead)  which could be problematic in some cases.

saadrahim:
@eidenyoshida you may want to finish this off.

eidenyoshida:
Opened a separate PR to add sanitizer flags to install script

PR #163, opened 2021-02-16
Add address sanitizer option to install script

eidenyoshida:


eidenyoshida:
Need to run this on CI against a newer mainline build. 6132 is likely too old

.jenkins/common.groovy
```diff
@@ -27,14 +27,14 @@ def runTestCommand (platform, project, boolean debug=false)
     String sudo = auxiliary.sudo(platform.jenkinsLabel)
     String centos = platform.jenkinsLabel.contains('centos') ? '3' : ''
     String buildTypeDir = debug ? 'debug' : 'release'
-    String sanitizerLibPath = debug ? "export LD_LIBRARY_PATH=/opt/rocm/llvm/lib/clang/12.0.0/lib/linux:\$LD_LIBRARY_PATH" : ''
+    String sanitizerLibPath = debug ? "/opt/rocm/llvm/lib/clang/12.0.0/lib/linux" : ''
```
saadrahim, line 30: I wish we have a more upgrade proof way to add this.

saadrahim: define it in rocProject?

install
```diff
@@ -89,6 +94,9 @@ while true; do
     --hip-clang)
         build_hip_clang=true
         shift ;;
+    --sanitizer)
```
jsandham, line 97: If using sanitizer does it need to be built in debug? I.e. should build_release=false be set when using --sanitizer?

eidenyoshida: Yes it makes sense to build in debug mode, however I'm not sure if it is required

jsandham:
I was looking at adding this for rocSPARSE. Do you think it would be better to instead do something like:

# address sanitizer
if [[ "${build_sanitizer}" == true ]]; then
  cmake_common_options="${cmake_common_options} -DBUILD_ADDRESS_SANITIZER=ON"
fi

in the install script and then in the CMakeList.txt:

# Build options
option(BUILD_SHARED_LIBS "Build rocSPARSE as a shared library" ON)
option(BUILD_CLIENTS_TESTS "Build tests (requires googletest)" OFF)
option(BUILD_CLIENTS_BENCHMARKS "Build benchmarks (requires boost)" OFF)
option(BUILD_CLIENTS_SAMPLES "Build examples" ON)
option(BUILD_VERBOSE "Output additional build information" OFF)
option(BUILD_CODE_COVERAGE "Build rocSPARSE with code coverage enabled" OFF)
option(BUILD_ADDRESS_SANITIZER "Build rocSPARSE with address sanitizer enabled" OFF)

if(BUILD_ADDRESS_SANITIZER)
  add_compile_options(-g -fsanitize=address -shared-libasan)
  add_link_options(-g -fuse-ld=lld)
endif()

This way its easy to set when not using the install script (which I have heard may be replaced or not used at some point)? This is the way we added code coverage. Thoughts?

eidenyoshida:
That seems like a good plan to me. A cmake approach is easier to implement across all the libraries

eidenyoshida:
> I was looking at adding this for rocSPARSE. Do you think it would be better to instead do something like:
> 
> // address sanitizer
> if [[ "${build_sanitizer}" == true ]]; then
> cmake_common_options="${cmake_common_options} -DBUILD_ADDRESS_SANITIZER=ON"
> fi
> 
> in the install script and then in the CMakeList.txt:
> 
> // Build options
> option(BUILD_SHARED_LIBS "Build rocSPARSE as a shared library" ON)
> option(BUILD_CLIENTS_TESTS "Build tests (requires googletest)" OFF)
> option(BUILD_CLIENTS_BENCHMARKS "Build benchmarks (requires boost)" OFF)
> option(BUILD_CLIENTS_SAMPLES "Build examples" ON)
> option(BUILD_VERBOSE "Output additional build information" OFF)
> option(BUILD_CODE_COVERAGE "Build rocSPARSE with code coverage enabled" OFF)
> option(BUILD_ADDRESS_SANITIZER "Build rocSPARSE with address sanitizer enabled" OFF)
> 
> if(BUILD_ADDRESS_SANITIZER)
> add_compile_options(-g -fsanitize=address -shared-libasan)
> add_link_options(-g -fuse-ld=lld)
> endif()
> 
> This way its easy to set when not using the install script (which I have heard may be replaced or not used at some point)? This is the way we added code coverage. Thoughts?
> 
> Building without the install script would then work like:
> 
> CXX=/opt/rocm/bin/hipcc cmake -DBUILD_CLIENTS_TESTS=ON -DBUILD_ADDRESS_SANITIZER=ON ../..
> make
> make install

is -g preferred or is setting CMAKE_BUILD_TYPE=Debug the way to go?

jsandham:
Is -fuse-ld=lld required? Without this I think it will just use the hipcc linker which should be fine?

.jenkins/common.groovy
```diff
@@ -10,37 +10,39 @@ def runCompileCommand(platform, project, jobName, boolean debug=false, boolean s
...
     String buildStatic = staticLibrary ? '-DBUILD_STATIC_LIBS=ON' : '-DBUILD_SHARED=OFF'
     String cmake = platform.jenkinsLabel.contains('centos') ? 'cmake3' : 'cmake'
-
+    String addressSanitizerFlags = addressSanitizer ? "CFLAGS='-fsanitize=address -shared-libasan' CXXFLAGS='-fsanitize=address -shared-libasan' LDFLAGS='-fuse-ld=lld'": ""
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}
                 mkdir -p build/${buildTypeDir} && cd build/${buildTypeDir}
-                ${cmake} -DCMAKE_CXX_COMPILER=/opt/rocm/bin/hipcc ${buildTypeArg} ${buildStatic} -DBUILD_TEST=ON -DBUILD_BENCHMARK=ON ../..
+                ${addressSanitizerFlags} ${cmake} -DCMAKE_C_COMPILER=/opt/ropcm/bin/hipcc -DCMAKE_CXX_COMPILER=/opt/rocm/bin/hipcc ${buildTypeArg} ${buildStatic} -DBUILD_TEST=ON -DBUILD_BENCHMARK=ON ../..
```
ex-rzr, line 18: typo: ropcm

PR #165, opened 2021-02-22
Workaround performance fix for xorwow

stanleytsang-amd:
The workaround restores lost performance for xorwow engine in newer ROCm builds.  This affected both MI50/Vega20 and MI100.

I'll file an internal ticket with the HIP compiler team shortly about these regressions.

### DIFF
--- a/library/include/rocrand_xorwow.h
+++ b/library/include/rocrand_xorwow.h
@@ -78,9 +78,6 @@ class xorwow_engine
 public:
     struct xorwow_state
     {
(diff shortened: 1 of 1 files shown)


stanleytsang-amd:
It looks like the performance regression for xorwow has been fixed in ROCm 4.1 RC2 and newer mainline builds.  This workaround has no effect in those builds, but is still valid for older builds.

neon60 (approved):

PR #174, opened 2021-05-03
Sobol 64

neon60:
Sobol 64 without optimization


cmake/VerifyCompiler.cmake
```diff
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2018-2020 Advanced Micro Devices, Inc. All rights reserved.
```
doctorcolinsmith, line 3: 2021

library/src/rng/sobol64.hpp
```diff
@@ -0,0 +1,314 @@
+// Copyright (c) 2017 Advanced Micro Devices, Inc. All rights reserved.
```
doctorcolinsmith, line 1: 2021

doctorcolinsmith:
I noticed a few copyright statements that should be updated to 2021.  Otherwise, this looks good.

doctorcolinsmith (approved):

PR #176, opened 2021-05-10
Revert "Develop stream 20210414"

saadrahim:
Reverts ROCmSoftwarePlatform/rocRAND#173

This is changing how hipRAND is packaged. I am reverting this PR immediately.


neon60:
Hi @saadrahim 

Can you share the problem with the hipRAND package?

PR #186, opened 2021-06-23
Add hmm tests to gfx90a CI runs

eidenyoshida:


.jenkins/common.groovy
```diff
@@ -31,12 +31,22 @@ def runTestCommand (platform, project)
...
+                         """
+    }
 
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}/build/release
                 make -j4
-                ${sudo} LD_LIBRARY_PATH=/opt/rocm/lib/ ${testCommand}
+                ${testCommand}
+                ${hmmTestCommand}
```
doctorcolinsmith, line 49: `testCommand` is included in `hmmTestCommand` already.  Is intentional to include both in `command`?

eidenyoshida: Yes, the idea is to run the full suite of tests on all gfx architectures. On gfx90a, we would repeat the suite but with hmm and xnack enabled.

stanleytsang-amd (approved):

eidenyoshida:
Will reopen when new build available

PR #190, opened 2021-07-26
HIP on Windows initial support

stanleytsang-amd:
Initial HIP on Windows support for rocRAND. Currently on my test machine rocRAND builds, installs, and unit tests pass.

How to build with HIP on Windows:

Get system dependencies with rdeps.py: python rdeps.py
Install using new rmake.py script: python rmake.py
To run tests: python rtest.py -t <insert test set name here, see rtest.xml for test set names>

I have not tested rmake.py with linux yet, I suspect it still needs some work, but the currently install.sh script works just as it did before for linux. Long-term we can decide what we want to do with install.sh/rmake.py.


CMakeLists.txt
```diff
@@ -23,11 +23,19 @@
 cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 
 # Install prefix
-set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories")
-
-#Adding CMAKE_PREFIX_PATH
-list( APPEND CMAKE_PREFIX_PATH /opt/rocm/llvm /opt/rocm )
+if (WIN32)
+  SET( CMAKE_INSTALL_PREFIX "C:/hipSDK" CACHE PATH "Install path" FORCE )
```
saadrahim, line 27: @TorreZuk when do we change this into something more appropriate for a windows install path?

TorreZuk: > @TorreZuk when do we change this into something more appropriate for a windows install path?

@saadrahim I thought you were organizing the packaging people.  But didn't you mention last week someone was assigned now?   After our next meeting if there is clarity it can be tasked up then.  For now the libs I expect are installing into C:/hipSDK.
An SDK installer builder can likely make a package/msi from contents in hipSDK and put them in a new tree, and merge with C:/hip or not merge...   This is fine for now as matches the other libs more or less.   Also the symlink duplication weren't you looking into the preferred solution?

stanleytsang-amd:
Test failure on gfx90a is a known failure and is not related to any of the HIP on Windows changes.

CMakeLists.txt
saadrahim: It is slow progress with packaging. I think we should consider using an environment variable like HIP_PATH or equivalent for this in the future in our code instead of  hard coding it.

TorreZuk: Yes that makes sense to use a variable, wasn't some document suggesting not to use environment variables, but anyway my mistake for not going that route from the start but let's leave them all until after the release and then switch to use one and set the desired default for the installations.    The only problem with injecting into the folder along with hip is the cleanup afterwards as not sure what tool can remove the contents of a zip from a folder.   And do we keep the common bin and include as done now?   Is there a ticket for the packaging and install ?

saadrahim (changes requested):
Please add to the changelog.

CMakeLists.txt
```diff
@@ -79,10 +88,27 @@ else()
...
 
 # Verify that hcc compiler is used on ROCM platform
-include(cmake/VerifyCompiler.cmake)
+# TODO: Fix VerifyCompiler for Windows
+if (NOT WIN32)
+  include(cmake/VerifyCompiler.cmake)
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH ${ROCM_PATH} ${ROCM_PATH}/hip ${ROCM_PATH}/llvm /opt/rocm/llvm /opt/rocm/hip)
+find_package(hip REQUIRED CONFIG PATHS ${HIP_DIR} ${ROCM_PATH} /opt/rocm)
```
neon60, line 111: These two lines are also in VerifyCompiler.cmake. Can you please move this to else statement:
```
if (NOT WIN32)
  include(cmake/VerifyCompiler.cmake)
else()
  list(APPEND CMAKE_PREFIX_PATH ${ROCM_PATH} ${ROCM_PATH}/hip ${ROCM_PATH}/llvm /opt/rocm/llvm /opt/rocm/hip)
  find_package(hip REQUIRED CONFIG PATHS ${HIP_DIR} ${ROCM_PATH} /opt/rocm)
endif()
```

neon60:
Please check the package test on windows:
https://github.com/ROCmSoftwarePlatform/rocRAND/blob/develop/test/package/CMakeLists.txt

stanleytsang-amd:
> Please add to the changelog.

done

README.md
```diff
@@ -32,6 +32,9 @@ environment cuRAND is used instead.
 * For CUDA platforms:
   * [HIP](https://github.com/ROCm-Developer-Tools/HIP) (hcc is not required)
   * Latest CUDA SDK
+* Python 3.6 or higher (HIP on Windows only)
```
saadrahim, line 35: not really needed, convenience layer.

saadrahim (approved):

PR #193, opened 2021-07-27
Add regex override to rtest.xml

eidenyoshida:


lawruble13 (approved):
LGTM. For posterity, note that versions of rtest.py which don't recognize the `failure-regex` attribute will simply ignore it, which may cause spurious failures. If `failure-regex` is not specified it defaults to `error|fail`, and is not case sensitive.

saadrahim (approved):

PR #194, opened 2021-07-29
Fix runtime package postinst & prerm

lawruble13:
Runtime package did not create `${inc_dir}`, so symlinking to `${inc_dir}/../../<path>` fails on fresh installation of runtime package.


saadrahim (approved):

saadrahim:
@stanleytsang-amd this needs to go into master ASAP

PR #240, opened 2022-01-10
Change install script hipRAND download behaviour

lawruble13:
Install script will download hipRAND if
- it is trying to build hipRAND (which it is by default), and
- hipRAND/CMakeLists.txt does not exist.


stanleytsang-amd:
@lawruble13 Do we need to update rmake.py for HIP on Windows as well?

lawruble13:
HIP is currently building hipRAND separately, and builds hipRAND separately in the CMake as well, so it's probably not required.

stanleytsang-amd (approved):

PR #244, opened 2022-01-21
Hotfix: symlinks for compatibility

lawruble13:
Create symlinks in the old rocrand header file locations for compatibility.


saadrahim (approved):
If CI passes, I have no concerns.

PR #266, opened 2022-04-07
Add CMake HIP language support

cgmb:
## Build Instructions
1. Install ROCm 4.5.2
2. Install CMake 3.21.3 or later
3. Install libnuma and kmod (so that `rocm_agent_enumerator` works)
4. Build rocRAND:
```bash
HIPCXX=/opt/rocm/llvm/bin/clang++ \
CXX=/opt/rocm/llvm/bin/clang++ \
CC=/opt/rocm/llvm/bin/clang \
cmake -S. -Bbuild -GNinja \
  -DBUILD_HIPRAND=OFF \
  -DBUILD_FILE_REORG_BACKWARD_COMPATIBILITY=OFF \
  -DROCRAND_CMAKE_HIP_LANGUAGE=ON
```

With these changes, the above command will build a version of the library compatible with the system it's running on. If you want specific GPU architectures, append `-DCMAKE_HIP_ARCHITECTURES='gfx906:xnack-;gfx1030'` (replacing the example `gfx906:xnack-;gfx1030` values with your chosen semi-colon delimited list of [AMDGPU target ids](https://llvm.org/docs/AMDGPUUsage.html#processors)).

## Related CMake Documentation
- [CMAKE_HIP_ARCHITECTURES](https://cmake.org/cmake/help/v3.23/variable/CMAKE_HIP_ARCHITECTURES.html)
- [CMAKE_HIP_STANDARD](https://cmake.org/cmake/help/v3.23/prop_tgt/CMAKE_HIP_STANDARD.html)
- [HIPCXX](https://cmake.org/cmake/help/v3.23/envvar/HIPCXX.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,15 +69,6 @@ include(cmake/Dependencies.cmake)
 option(ROCRAND_CMAKE_HIP_LANGUAGE "Use CMake's HIP language support" OFF)
 if(ROCRAND_CMAKE_HIP_LANGUAGE)
   enable_language(HIP)
-  set(CMAKE_HIP_ARCHITECTURES
-    gfx803
-    gfx900:xnack-
-    gfx906:xnack-
-    gfx908:xnack-
-    gfx90a:xnack-
-    gfx90a:xnack+
-    gfx1030
-  )
 else()
   # Detect compiler support for target ID
   # This section is deprecated. Please use rocm_check_target_ids for future use.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2021 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2018-2022 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -66,39 +66,44 @@ set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared")
...
-if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
-    execute_process(COMMAND ${CMAKE_CXX_COMPILER} "--help"
-        OUTPUT_VARIABLE CXX_OUTPUT
-        OUTPUT_STRIP_TRAILING_WHITESPACE
-        ERROR_STRIP_TRAILING_WHITESPACE)
-    string(REGEX MATCH ".mcode\-object\-version" TARGET_ID_SUPPORT ${CXX_OUTPUT})
-endif()
+option(ROCRAND_CMAKE_HIP_LANGUAGE "Use CMake's HIP language support" OFF)
+if(ROCRAND_CMAKE_HIP_LANGUAGE)
+  enable_language(HIP)
```
pfultz2, line 71: You still need to call `find_package(hip)` when using hip language. Libraries like rocRAND need to provide `hip::host` in the targets interface because it includes hip's runtime API headers.

cgmb: Thanks. I was just about to request your review. I haven't yet tested this with downstream libraries.

pfultz2: Ah wait it looks like rocRAND actually needs hip::device as it includes hip's device headers:

https://github.com/ROCmSoftwarePlatform/rocRAND/blob/develop/library/include/rocrand/rocrand.h#L29

We would need two different targets to support using rocRAND with HIP language and with CXX language.

cgmb: It's not actually clear to me why it includes `<hip/hip_device.h>`. rocRAND appears to provide a C99 API. Sadly, [the documentation](https://rocrand.readthedocs.io/) is... lacking.

cgmb: > We would need two different targets to support using rocRAND with HIP language and with CXX language.

Maybe we could detect the language with a generator expression? I'm not entirely sure it's possible, but I was thinking something like:
```cmake
set_target_properties(roc::rocrand
  PROPERTIES
    INTERFACE_LINK_LIBRARIES "$<$<NOT:$<COMPILE_LANGUAGE:HIP>>:hip::device>"
)
```

test/CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ endfunction()
 
 # Get rocRAND tests source files
 file(GLOB rocRAND_TEST_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
+if(ROCRAND_CMAKE_HIP_LANGUAGE)
+    set_source_files_properties(${rocRAND_TEST_SRCS}
+      PROPERTIES
+        LANGUAGE HIP
+    )
+endif()
```
Maetveis, line 27: It would be better to have these lines together with the lines for CUDA a few lines below.

benchmark/CMakeLists.txt
```diff
@@ -29,7 +34,7 @@ foreach(benchmark_src ${rocRAND_BENCHMARK_SRCS})
     else()
         target_link_libraries(${BENCHMARK_TARGET}
             rocrand
-            hip::device
+            $<$<NOT:$<BOOL:ROCRAND_CMAKE_HIP_LANGUAGE>>:hip::device>
```
Maetveis, line 37: Nit: for benchmarks an if statement is used, but here its a generator expression. I would stick to the if in both places for consistency, and because this setting is known at configure time, a generator expression is not required.

Maetveis:
HI! rocRAND developer here, thank you for getting this off the ground. I believe cmake HIP support for rocRAND would simplify many projects that depend on rocRAND. I added a few comments, and will try this out soon.

I think I can help answer some questions regarding the project if you need anything.

CMakeLists.txt
```diff
@@ -66,39 +66,44 @@ set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared")
...
-# This section is deprecated. Please use rocm_check_target_ids for future use.
-if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
-    execute_process(COMMAND ${CMAKE_CXX_COMPILER} "--help"
-        OUTPUT_VARIABLE CXX_OUTPUT
-        OUTPUT_STRIP_TRAILING_WHITESPACE
-        ERROR_STRIP_TRAILING_WHITESPACE)
-    string(REGEX MATCH ".mcode\-object\-version" TARGET_ID_SUPPORT ${CXX_OUTPUT})
-endif()
+option(ROCRAND_CMAKE_HIP_LANGUAGE "Use CMake's HIP language support" OFF)
+if(ROCRAND_CMAKE_HIP_LANGUAGE)
```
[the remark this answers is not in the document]
cgmb: That's an interesting thought. I'll change this behaviour from `ROCRAND_CMAKE_HIP_LANGUAGE=ON` to `ROCRAND_CMAKE_LANG=HIP`. The default would be `ROCRAND_CMAKE_LANG=CXX`, and if somebody wants to implement CMake CUDA language support, it would be `ROCRAND_CMAKE_LANG=CUDA`.

library/CMakeLists.txt
```diff
@@ -48,11 +48,18 @@ if(HIP_COMPILER STREQUAL "nvcc")
...
+          PROPERTIES
+            LANGUAGE HIP
+        )
+    else()
+        target_link_libraries(rocrand PRIVATE hip::device)
+        if(NOT WIN32)
+            foreach(amdgpu_target ${AMDGPU_TARGETS})
+                target_link_libraries(rocrand PRIVATE --amdgpu-target=${amdgpu_target})
+            endforeach()
+        endif()
```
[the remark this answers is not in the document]
cgmb: I don't want to change any existing behaviour in this PR.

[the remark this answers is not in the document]
cgmb: I should add, I think it would be great to simplify the CMake and I fully support removing that code. Just not in this particular PR. I'd like to keep the addition of the new option and changes in the behaviour of existing options separate.

saadrahim:
This has been sitting around for a while. Should it be resurrected?

cgmb:
> This has been sitting around for a while. Should it be resurrected?

Maybe, but I don't have the bandwidth to do that quite yet.

The official Linux and Windows builds now both use CMake 3.22, so if this PR were finished, we could begin to use the CMake HIP language support in our official builds.

haampie:
@cgmb I'm trying to move the Spack ecosystem to use HIP as a language. This PR would be useful. Is it possible to get it in?

cgmb:
@haampie, that would be useful for Debian too. It would make it a lot easier to set common flags and options for all HIP-based libraries that way. I think the approach in this PR needs to be adjusted, but, yeah, I think I'll try to get this moving again...

jayhawk-commits:

cgmb:
I think this is still a feature that is desired, but this PR is so outdated that we should start a new one.

PR #267, opened 2022-04-21
Fixing benchmark and unit test build for Windows

stanleytsang-amd:
1. Copying necessary DLLs for benchmarks and unit tests (install (TARGETS rocrand DESTINATION "${PROJECT_BINARY_DIR}/test") does not seem to work)
2. Add required function overload in Windows


lawruble13 (approved):
LGTM

PR #271, opened 2022-05-16
avoids req. ROCM_PATH env var

TorreZuk:
* avoids use of env var. ROCM_PATH for other purpose

### DIFF
--- a/rmake.py
+++ b/rmake.py
@@ -88,7 +88,7 @@ def config_cmd():
     
     if (OS_info["ID"] == 'windows'):
         # we don't have ROCM on windows but have hip, ROCM can be downloaded if required
-        rocm_path = os.getenv( 'ROCM_PATH', "C:/github/rocm-cmake-master") #C:/hip") # rocm/Utils/cmake-rocm4.2.0"
+        rocm_path = os.getenv( 'ROCM_CMAKE_PATH', "C:/github/rocm-cmake-master") #C:/hip") # rocm/Utils/cmake-rocm4.2.0"
         cmake_executable = "cmake.exe"
         toolchain = os.path.join( src_path, "toolchain-windows.cmake" )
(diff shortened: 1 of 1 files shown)


DanyiLin:
The change works for rocRAND windows build.  Same question as @cgmb

TorreZuk:
Not sure what the "meta" failure is but must be unrelated.   Setting an environment variable is visible by scripts and processes so fools some compiler detection steps into thinking it is a linux-like rocm install at that location and so looks for libs there.   This variable was only be used only as a crutch for find_package(ROCM) which is local to cmake, i.e. setting a local cmake variable ROCM_PATH with that path is safe if only used in the PATHS argument to cmake find_package(ROCM) (which contrary to what the name suggests isn't a true "rocm" just rocm-cmake) so the fact that nothing but rocm-cmake was there was fine as hip has its own path too.  We should remove this crutch with either a better name, or unifying the rocm-cmake into a "rocm" like folder but again can be handled with PREFIX path and/or give rocm-cmake a better name and typical cmake variable name for such.  That is what Cory will draft a proposal for.

stanleytsang-amd (approved):

stanleytsang-amd:
I'm fine with this as a stopgap measure until a full proper fix is proposed.
Interestingly enough I tried removing add_compile_options(-xhip) Cmake and it builds successfully.  Seems like the compiler has been updated such that we no longer need to explicitly have this option, which might solve our problem another way.

TorreZuk:
> I'm fine with this as a stopgap measure until a full proper fix is proposed. Interestingly enough I tried removing add_compile_options(-xhip) from cmakelists.txt and it builds successfully. Seems like the compiler has been updated such that we no longer need to explicitly have this option, which might solve our problem another way.

You may as well make the PR and remove things that aren't required, best to keep windows as close as possible to linux.

PR #274, opened 2022-06-13
Remove FATAL_ERROR when HIP_COMPILER is not clang

cgmb:
When `HIP_COMPILER` is "clang-14", rocRAND emits a fatal error, "On ROCm platform 'hcc' or 'clang' must be used as C++ compiler" [[1]](https://lists.debian.org/debian-ai/2022/06/msg00002.html). As far as I can tell, these compiler checks are not necessary.


Maetveis:
Hi, although I am not an expert in HIP packaging, to me it seems like `HIP_COMPILER` should not be set to specific versions of the clang binary, instead judging from [how it's set in hipamd](https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/CMakeLists.txt#L222), and how its [used in `hip-config.cmake`](https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/hip-config.cmake.in#L126) it should be either just `clang` or `nvcc` (HIP atop CUDA).
In other libraries `HIP_COMPILER` (e.g hipCUB) is used to differentiate between targeting CUDA, or ROCm, in the same way as here.

I was surprised that the build did not fail with this setting, because lots of compiler flags from `hip-config.cmake`, would not be set up, until I found where [it is patched in debian](https://salsa.debian.org/rocm-team/rocm-hipamd/-/blob/master/debian/patches/0005-clang-14.patch). Mostly out of curiosity, can you tell me why `HIP_COMPILER` was also changed? I would think changing `HIP_CXX_COMPILER` (and paths) would be enough.

cgmb:
> Mostly out of curiosity, can you tell me why `HIP_COMPILER` was also changed? I would think changing `HIP_CXX_COMPILER` (and paths) would be enough.

I'm not sure. I will ask.

> In other libraries HIP_COMPILER (e.g hipCUB) is used to differentiate between targeting CUDA, or ROCm, in the same way as here.

Isn't that the purpose of `HIP_PLATFORM`?

Maetveis:
> Isn't that the purpose of `HIP_PLATFORM`?

find_package(HIP) (hip-config.cmake) does not set that, I think it is only reported by `hipconfig`, in code the macros `__HIP_PLATFORM_AMD__` and `__HIP_PLATFORM_NVIDIA__` are used, but these are set in `hip_runtime_api.h` i.e. cannot be used from cmake.

cgmb:
As far as I can tell, `HIP_COMPILER` is never actually used for compiling. That's... surprising, but I suppose it was just introduced to distinguish `hcc` and `clang`? Using it to distinguish AMD vs. NVIDIA seems like we're setting ourselves up for confusion, particularly given that there's CUDA support in clang.

I've suggested to Debian not to change the `HIP_COMPILER` and my (albeit limited) testing indicates that's working fine even when the actual binary doing the compilation is clang-14. The Debian HIP package is still a work-in-progress and it's honestly been pretty confusing, so I appreciate your help. Thanks for the advice!

PR #277, opened 2022-07-07
Googlebenchmark integration

Maetveis:
# rocRAND google benchmark integration

This pull request includes an initiative to replacing the current benchmark system with Google benchmark. Google benchmark has a number of advantages over the current benchmark system, like integrated options for filtering benchmarks by regex, and outputting machine-readable results in json or csv format. Furthermore, google benchmark is also used by a number of other ROCm repositories, such as rocPRIM and hipCUB, and so this reduces the total amount of custom benchmark processing scripts required.

The changes in this pull request have the following implications:
- Google benchmark is automatically downloaded during the build process if required, similar to google test and similar to the other repositories.
- The amount of parameter-handling code required is minimized, this is now handled by the `--benchmark_filter` option of Google benchmark.
- Some additional information about the benchmark is gathered and included in the benchmark output. This mostly consists of the `hipDeviceProp_t` of the device that the benchmarks were executed on, and can be useful during futher processing of the results.

## Additional changes
There are some minor changes included mainly addressing the gitlab CI
- the URL of the hipRAND submodule is changed to be relative going back only one level (removing the explicit ROCmSoftwarePlatform part) so the submodule works in a fork of rocRAND as long as hipRAND is also forked
- Updated .clang-format file and commit hooks for rejecting incorrectly formatted commits
- A workaround for the packaging test for the hip headers on the NVIDIA platform. (It is fixed in master but not yet in the latest rocm release)


cmake/Dependencies.cmake
```diff
@@ -69,6 +69,44 @@ if(BUILD_TEST)
...
+    find_package(benchmark QUIET)
+  endif()
+
+  if(NOT benchmark_FOUND)
+    message(STATUS "Google Benchmark not found or force download Google Benchmark on. Downloading and building Google Benchmark.")
+    if(CMAKE_CONFIGURATION_TYPES)
+      message(FATAL_ERROR "DownloadProject.cmake doesn't support multi-configuration generators.")
+    endif()
+    set(GOOGLEBENCHMARK_ROOT ${CMAKE_CURRENT_BINARY_DIR}/deps/googlebenchmark CACHE PATH "")
+    if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
```
stanleytsang-amd, line 86: For HIP on Windows, setting the compiler override to g++ breaks the build.  Can we preserve the old behaviour for HIP on Windows? ie.
if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (WIN32 AND CMAKE_CXX_COMPILER MATCHES ".*/hipcc$"))

cmake/Dependencies.cmake
```diff
@@ -69,6 +69,44 @@ if(BUILD_TEST)
...
+      # hip-clang cannot compile googlebenchmark for some reason
+      set(COMPILER_OVERRIDE "-DCMAKE_CXX_COMPILER=g++")
+    endif()
+
+    download_project(
+      PROJ           googlebenchmark
+      GIT_REPOSITORY https://github.com/google/benchmark.git
+      GIT_TAG        v1.6.1
+      INSTALL_DIR    ${GOOGLEBENCHMARK_ROOT}
+      CMAKE_ARGS     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} -DBENCHMARK_ENABLE_TESTING=OFF -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> ${COMPILER_OVERRIDE}
```
stanleytsang-amd, line 96: For HIP on Windows -DCMAKE_CXX_STANDARD=14 seems to be required in order to successfully build googlebenchmark.

Maetveis:
Restored the legacy generate benchmark, and added an option to enable building them (on by default). The default can be changed to OFF in later versions, before the legacy benchmarks are removed.

If legacy benchmarks are enabled a deprecation warning is printed to warn users, it can be disabled by passing `-Wno-deprecated` as documented for [`CMAKE_WARN_DEPRECATED`](https://cmake.org/cmake/help/latest/variable/CMAKE_WARN_DEPRECATED.html#variable:CMAKE_WARN_DEPRECATED)

Additional changes:
The commit hook script is made more portable by using `env`

Maetveis:
Dropped the compiler workaround in Dependencies.cmake, as the cause is fixed in googlebenchmark since 1.4.5 (the PR already requires 1.6.1)

nolmoonen:
Rebased the branch.

stanleytsang-amd (approved):

PR #279, opened 2022-07-21
Benchmark warmup increase

stanleytsang-amd:
When running the generate benchmark on Navi21 with select distributions + engine types, we get data that fits in the L3 cache, generating artificially high benchmark scores vs. running the entire benchmark suite, which can lead to a source of confusion for end users.  As @Maetveis suggested, increasing the number of warmup iterations will stop this phenomenon from happening.


Maetveis (approved):
LGTM

stanleytsang-amd:
Unsuccessful check is due to CI-specific failure, not rocrand specific. Merging.

PR #281, opened 2022-08-04
Re-enable inline asm for gfx1030

stanleytsang-amd:
QA performance testing revealed that enable inline assembly actually benefits navi21 in the rocRAND philox benchmarks (the previous assessment of inline asm being bad for MI200 is still correct though).

rocrand-kernel/philox-poisson: 14.25% loss without inline asm
rocrand-kernel/philox-discrete-custom: 5.89% loss
rocrand-kernel/philox-discrete-poisson: 5.52% loss
rocrand-generate/philox-poisson: 7.68% loss

### DIFF
--- a/library/include/rocrand/rocrand_common.h
+++ b/library/include/rocrand/rocrand_common.h
@@ -63,7 +63,8 @@ namespace detail {
       defined(__gfx904__) || \
       defined(__gfx906__) || \
       defined(__gfx908__) || \
-      defined(__gfx909__) )
+      defined(__gfx909__) || \
+      defined(__gfx1030__) )
   #if !defined(ROCRAND_ENABLE_INLINE_ASM)
     #define ROCRAND_ENABLE_INLINE_ASM
   #endif


arsenm:
Inline asm should never be beneficial. There should be a compiler bug if it is

stanleytsang-amd:
> Inline asm should never be beneficial. There should be a compiler bug if it is

@Maetveis @neon60 https://github.com/ROCmSoftwarePlatform/rocRAND/blob/develop/library/include/rocrand/rocrand_common.h#L78 is the only part of rocRAND which potentially uses inline asm, right?  Can you provide some background on this?

In any case, I'm all for merging this while we file a compiler bug once we get more info.

doctorcolinsmith (approved):

stanleytsang-amd:
As discussed in weekly meeting, StreamHPC will investigate and create a minimal repro which can be sent to compiler team for analysis as to why inline asm seems to result in better performance in this case.  In the meantime we'll merge this in.

searlmc1:
Hi @stanleytsang-amd ,

> As discussed in weekly meeting, StreamHPC will investigate and create a minimal repro which can be sent to compiler team for analysis as to why inline asm seems to result in better performance in this case. In the meantime we'll merge this in.

Any update on this?

Thanks

stanleytsang-amd:
> Hi @stanleytsang-amd ,
> 
> > As discussed in weekly meeting, StreamHPC will investigate and create a minimal repro which can be sent to compiler team for analysis as to why inline asm seems to result in better performance in this case. In the meantime we'll merge this in.
> 
> Any update on this?
> 
> Thanks

Hi Mark, I've just created an internal ticket (SWDEV-356009) to the compiler team with a minimal repro and explanation.  You should be a watcher on it.

searlmc1:
> Hi @searlmc1, I've just created an internal ticket (SWDEV-356009) to the compiler team with a minimal repro and explanation. You should be a watcher on it.

Thanks!

searlmc1:
> > Hi @searlmc1, I've just created an internal ticket (SWDEV-356009) to the compiler team with a minimal repro and explanation. You should be a watcher on it.

https://reviews.llvm.org/D133768 landed for this

PR #282, opened 2022-08-05
Fixes to enable googlebenchmark build on Windows

stanleytsang-amd:
As per my comments in https://github.com/ROCmSoftwarePlatform/rocRAND/pull/277.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -83,13 +83,17 @@ if(BUILD_BENCHMARK)
       message(FATAL_ERROR "DownloadProject.cmake doesn't support multi-configuration generators.")
     endif()
     set(GOOGLEBENCHMARK_ROOT ${CMAKE_CURRENT_BINARY_DIR}/deps/googlebenchmark CACHE PATH "")
-
+    if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (WIN32 AND CMAKE_CXX_COMPILER MATCHES ".*/hipcc$"))
+      # hip-clang cannot compile googlebenchmark for some reason
+      set(COMPILER_OVERRIDE "-DCMAKE_CXX_COMPILER=g++")
+    endif()
+	
     download_project(
       PROJ           googlebenchmark
       GIT_REPOSITORY https://github.com/google/benchmark.git
       GIT_TAG        v1.6.1
       INSTALL_DIR    ${GOOGLEBENCHMARK_ROOT}
-      CMAKE_ARGS     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} -DBENCHMARK_ENABLE_TESTING=OFF -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> ${COMPILER_OVERRIDE}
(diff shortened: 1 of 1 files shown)


cmake/Dependencies.cmake
```diff
@@ -83,13 +83,17 @@ if(BUILD_BENCHMARK)
       message(FATAL_ERROR "DownloadProject.cmake doesn't support multi-configuration generators.")
     endif()
     set(GOOGLEBENCHMARK_ROOT ${CMAKE_CURRENT_BINARY_DIR}/deps/googlebenchmark CACHE PATH "")
-
+    if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (WIN32 AND CMAKE_CXX_COMPILER MATCHES ".*/hipcc$"))
+      # hip-clang cannot compile googlebenchmark for some reason
+      set(COMPILER_OVERRIDE "-DCMAKE_CXX_COMPILER=g++")
+    endif()
```
Maetveis, line 89: I believe the issue that prevented compilation with clang on linux is fixed in googlebenchmark 1.5.4, since the PR already aims for 1.6.1 I think this is no longer necessarry.

cmake/Dependencies.cmake
```diff
@@ -83,13 +83,17 @@ if(BUILD_BENCHMARK)
...
+	
     download_project(
       PROJ           googlebenchmark
       GIT_REPOSITORY https://github.com/google/benchmark.git
       GIT_TAG        v1.6.1
       INSTALL_DIR    ${GOOGLEBENCHMARK_ROOT}
-      CMAKE_ARGS     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} -DBENCHMARK_ENABLE_TESTING=OFF -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> ${COMPILER_OVERRIDE}
+      CMAKE_ARGS     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} -DBENCHMARK_ENABLE_TESTING=OFF -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> -DCMAKE_CXX_STANDARD=14 ${COMPILER_OVERRIDE}
```
Maetveis, line 96: For me googlebenchmark still failed to configure on windows even with this change.
I created https://github.com/google/benchmark/issues/1462 for that error.
Are you sure that this is sufficient to make benchmark build successfully?

Maetveis (changes requested):

cmake/Dependencies.cmake
stanleytsang-amd: On my Windows test machine, yes.  The two changes in this PR seem to be required to get googlebenchmark to build successfully.  I wonder why this works for me and not for your environment - I assume you're using the HIP on Windows build Saad gave you?  Did it come with additional environment setup instructions?  Trying to figure out what might be causing the difference here.

Maetveis: > I assume you're using the HIP on Windows build Saad gave you?

Yes I am.

I just tried with cmake 3.16.9 and it works if `CMAKE_CXX_STANDARD` is set as you have here, so it seems like its the cmake version that's making the difference. I didn't try and bisect which is the last working version.

I am okay with this workaround until googlebenchmark is fixed.

PR #285, opened 2022-09-06
Create a valid CTestTestfile for installation

lawruble13:
Do not merge until hipRAND is merged & the reference is updated.


cgmb:
I wonder if we could get this feature into CMake itself. It would obviously take a while, so we'd definitely need our own way of doing it for now. Still, the ability to export and install test targets seems like something that could be done quite cleanly within CMake itself, and our own implementation might benefit from that design discussion with the CMake maintainers.

lawruble13:
@cgmb I definitely agree that this could be very cleanly done in CMake itself, and I was actually quite surprised to find that it didn't seem to be possible already. I think those would be good discussions to open, but in the interim does this solution seem acceptable?

cgmb:
Yes. I won't be able to able to examine your solution in detail, but it looks good to me.

lawruble13:
@stanleytsang-amd Can you take a look at this one as well? I needed to update the hipRAND reference, so it's running through all of its tests again.

stanleytsang-amd (approved):

littlewu2508:
Sorry, may I ask what is the purpose of installing cmake test file, especially into `bin/` directory? That cause distribution packages install it into `/usr/bin/rocRAND` which apparently breaks the FHS.

Note that the `test` component does not stop cmake from installing. Similar issue is https://github.com/RadeonOpenCompute/ROCm-CompilerSupport/issues/61

pfultz2:
The [ROCMTest](https://radeonopencompute.github.io/rocm-cmake/doc/html/ROCMTest.html#rocmtest) module will handle packaging tests and generating a CTestfile.cmake. Tests are a seperate component and will not be installed with `make install`, but `make install-tests` is provided to install the tests separately. 

The CTestfile.cmake is installed into `${CMAKE_INSTALL_PREFIX}/share/test/${NAME}` prefix, where `${NAME}` is the name of your component. So users can run `ctest ${CMAKE_INSTALL_PREFIX}/share/test/${NAME}` to run all the tests. All the executable binaries for tests are installed into `${CMAKE_INSTALL_PREFIX}/share/test/${NAME}/bin`, this is to avoid test executable names clashing across components. 

Of course, if `${CMAKE_INSTALL_PREFIX}/share/test/${NAME}` breaks FHS, then we can update the path in rocm-cmake, and if all components use the ROCMTest module then all components should get updated.

PR #294, opened 2022-09-23
Windows googlebench ci fix

stanleytsang-amd:
The current googlebenchmark integration PR is having a build failure with the Windows CI.  I changed the override compiler for windows to be the visual studio compiler - that seems to resolve the build failure in CI, see https://github.com/ROCmSoftwarePlatform/rocRAND/pull/293.  I wanted to run the change by you guys first to make sure this change doesn't break things on your Windows dev environment.  Once the fix is in we can finally merge in the googlebenchmark integration PR.


nolmoonen:
@stanleytsang-amd Thank you for giving us time to check this, I have confirmed that this change also works on our end.

PR #299, opened 2022-11-05
Update hip path to latest folder structure

arvindcheru:
Proposed Changes
- Updated module path, bin path to latest folder structure ($ROCM_PATH}/lib/cmake/hip, ${ROCM_PATH}/bin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ endif()
 list(APPEND CMAKE_MODULE_PATH
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules
+    $ENV{ROCM_PATH}/lib/cmake/hip
     ${HIP_PATH}/cmake $ENV{ROCM_PATH}/hip/cmake # FindHIP.cmake
     $ENV{ROCM_PATH}/llvm
 )
--- a/install
+++ b/install
@@ -185,7 +185,7 @@ if [[ "${build_relocatable}" == true ]]; then
     -DCMAKE_PREFIX_PATH="${rocm_path} ${rocm_path}/hcc ${rocm_path}/hip" \
     -DCMAKE_SHARED_LINKER_FLAGS="${rocm_rpath}" \
     -DROCM_DISABLE_LDCONFIG=ON \
-    -DCMAKE_MODULE_PATH="${rocm_path}/hip/cmake" \
+    -DCMAKE_MODULE_PATH="${rocm_path}/lib/cmake/hip ${rocm_path}/hip/cmake" \
     ../../. # or cmake-gui ../.
 else
     CXX=${rocm_path}/bin/$compiler ${cmake_executable} ${clients} ${cmake_common_options} ../../. # or cmake-gui ../.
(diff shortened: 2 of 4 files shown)


saadrahim:
@lawruble13 @arvindcheru should these paths be set manually in each project or should they be part of find_package(hip)?

arvindcheru:
> @lawruble13 @arvindcheru should these paths be set manually in each project or should they be part of find_package(hip)?

Existing code uses module path set (Currently Module mode searching HIP is done during rocRAND build) to old hip location, So this module path needs to be updated adding new path is proposed as part of this change.

cgmb:
> > @lawruble13 @arvindcheru should these paths be set manually in each project or should they be part of find_package(hip)?
> 
> Existing code uses module path set (Currently Module mode searching HIP is done during rocRAND build) to old hip location, So this module path needs to be updated adding new path is proposed as part of this change.

Setting the module path is only necessary because we use FindHIP.cmake, which we've been told repeatedly to stop doing. As I understand it, the HIP team considers it deprecated.

Maetveis:
> Setting the module path is only necessary because we use FindHIP.cmake, which we've been told repeatedly to stop doing. As I understand it, the HIP team considers it deprecated.

IIRC FindHIP.cmake is only used when targeting NVIDIA via HIP, and AFAIK hip-config.cmake doesn't work / isn't provided on NVIDIA platforms.

cgmb:
> IIRC FindHIP.cmake is only used when targeting NVIDIA via HIP, and AFAIK hip-config.cmake doesn't work / isn't provided on NVIDIA platforms.

Then SWDEV-243403 should be repoened.

saadrahim (dismissed):
Please fix the underlying issue as described by @cgmb

TorreZuk (commented):
I don't think @arvindcheru should be trying to fix or even change hip module support or hip-config consumption for nvcc use in this PR.   Aravindan is working on "Update hip path to latest folder structure".   
In fact I would **discourage** the change request, PRs need to remain focused for the original request and not get additions thrown in (best practice).  I suggest new ticket(s) should be made if you want changes hip consumption.

As to changes I quickly looked at the PR and I suggest minor fix but approve.

The top level CMakeLists.txt file I think should be improved by rocRAND team, verifyCompiler.cmake should not be doing the find (hip/HIP) that should be done at the top level code as linux/windows handling should both at same file.  If you want me to make a ticket for you I can, it relates to the other request (hip vs HIP).

cgmb:
> Aravindan is working on "Update hip path to latest folder structure".

That's a solution, not a problem. The problem is that `find_package(HIP)` no longer works the same way it did before, due to changes in the organization of the HIP package. There are multiple possible solutions to that problem. Adding the new location to the `CMAKE_MODULE_PATH` is one of those solutions.

The proper long-term solution is to not use module-mode search at all. Migrating the library to a config-mode search is more than what I could reasonably ask of @arvindcheru, but there should be a ticket in the backlog to do it right. We don't have to do it today or even tomorrow, but we've let this problem linger for over two years and we continue to pay a maintenace burden for that.

cgmb (approved):
@saadrahim, I don't think we want to thrust this problem on @arvindcheru. We will need to revisit the conditional on the `find_package` call anyway when we replace `BUILD_WITH_LIB`.

saadrahim:
@stanleytsang-amd I dropped my concerns as they were too broad. Please approve if you see fit.

arvindcheru:
Thank you for all the detailed comments. 
Will update the changes adding delimiter (;) for the paths as per the comments.

TorreZuk (approved):
Approve but not my repo, so you should still should get @stanleytsang-amd or other team member to approve if all your tests pass.  @cgmb you should make a new ticket or add your comments to the rejected ticket Saad referred to.

stanleytsang-amd (approved):

PR #301, opened 2022-11-18
Remove RPATH/RUNPATH from ROCm libraries and binaries

raramakr:
SWDEV-310152:
- Single version package: add ldconfig, no RPATH/RUNPATH in either binaries or libraries
- Multi version package: use RPATH for binaries, no RPATH/RUNPATH for libraries
- amdclang/clang/hipcc compilers shall not add RPATH/RUNPATH to produced binaries or libraries unless explicitly requested by an option
- Versioning scripts will take care of adding rpath to binaries for multi version package


stanleytsang-amd:
This PR is stale.  Closing.

PR #302, opened 2022-11-23
Copy googlebenchmark DLL to rocRAND benchmark build folder

stanleytsang-amd:
Googlebenchmark's benchmark*.dll is required to be in the same folder as the rocRAND benchmark executable on Windows.   This fixes benchmark_rocrand_host_api from throwing a runtime error about the missing dll.


SergiyKostrov:
> 

There is an env-variable GOOGLEBENCHMARK_ROOT to a folder with required DLLs. The script could SET ( or export in Linux terms ) that env-variable during benchmarks of rocRAND.

Snektron:
When installing the rocRAND benchmarks alongside the main library, this would not also copy the dll and so running those benchmarks would still yield the same problem. Copying the DLL to the bin folder could cause compatibility issues with other libraries when they are also modified to have this change.

An alternative solution that does not suffer from that problem might be to simply enforce to build googlebenchmark as a static library, by changing the `-DBUILD_SHARED_LIB=${BUILD_SHARED_LIB}` to `-DBUILD_SHARED_LIB=OFF` here:
https://github.com/ROCmSoftwarePlatform/rocRAND/blob/7d1f273030d9dda531cc403d0bd2bfa581c38482/cmake/Dependencies.cmake#L100

cgmb:
> Googlebenchmark's benchmark*.dll is required to be in the same folder as the rocRAND benchmark executable on Windows.

Note that the DLL copying solution was just the approach we took when we ran out of time to come up with something better. At some point, we should probably revisit that approach. 

For most DLLs, we could just set the PATH to add the library directory to the DLL search paths. [Copying DLLs is only really necessary to override a DLL that's installed in a system directory](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications), like libhipamd64.dll (which is installed with the drivers). I suppose it's also necessary if we want to build `*.zip` packages for the tests that include all dependencies.

If we do want to use bundling with DLLs, I think the 'right' way is to use something like [`install(IMPORTED_RUNTIME_ARTIFACTS targets...)`](https://cmake.org/cmake/help/v3.25/command/install.html#imported-runtime-artifacts). Though we would have to check what minimum CMake version we can require on Windows for building the libraries.

As mentioned by Robin, using static libraries would also be another possibility. That's how the Linux packaging ensures all the test dependencies are included, so it seems like a reasonable thing to do. I suppose we'd have to check if that approach would have any drawbacks on Windows, but I don't see any issues in principle.

Snektron:
It seems that in rocPRIM googlebenchmark is also linked statically, though its fetched via FetchContent rather than via the download script.

SergiyKostrov:
[AMD Official Use Only - General]

>>... rocPRIM googlebenchmark is also linked statically...

This is simply my point of view:

A DLL should be benchmarked or tested without embedding some additional functionality, for example using static linking of Google Test or Google Benchmark APIs. The DLL should be tested as is. It is possible that the test and performance data could be skewed.

stanleytsang-amd:
Thanks for the feedback all.  Agreed that copying DLL's is not ideal.  I will go about refactoring rocRAND to match rocPRIM's behaviour (use FetchContent, and link statically)

stanleytsang-amd:
For now I've changed googlebenchmark to build as a static library as we need to get a quick fix in for ROCm 4.5.2.  I had tried using FetchContent like we do in rocPRIM but came across various errors and will likely take too much effort to get this in for 4.5.2.

doctorcolinsmith (approved):

PR #307, opened 2022-12-22
Fix Threefry4x32 and Threefry4x64 array indexing

nolmoonen:
Corrects the array indexing for the Threefry4x32 and Threefry4x64 generators. This bug caused incorrect sequences to be produced by these two generators.


stanleytsang-amd:
All unit tests pass. CI failure is unrelated.

stanleytsang-amd (approved):

PR #311, opened 2023-01-24
Workaround for ROCm 5.5 performance regression

Snektron:
This is a fix for a performance regression observed when compiling rocRAND legacy benchmarks using a recent version of rocm-llvm. This regression is caused by commit ba0d079c7aa52bc0ae860d16dd4a33b0dc5cfff7 in upstream llvm, and just so happens to cause problems for this kernel. This workaround reverts to the old way that the stride is computed, however, this should in theory be less efficient than using gridDim and blockDim.

### DIFF
--- a/benchmark/benchmark_rocrand_kernel.cpp
+++ b/benchmark/benchmark_rocrand_kernel.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (c) 2017-2023 Advanced Micro Devices, Inc. All rights reserved.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


Snektron:
It looks like this issue is already fixed with the current version of `amd-stg-open`. I think one of the commits that was pulled in in the last few days has fixed it, but if the llvm version is already tagged for ROCm 5.5, then this fix is still required.

arsenm (commented):
This is a really unstructured workaround that doesn't make any sense

Snektron:
Indeed. I believe that the change to the scheduler in that LLVM commit just so happens to cause a regression in the specific code used in this benchmark: it looks like that pass gets activated when the register usage exceeds a lower- and upper bound, and using `gridDim.x * blockDim.x` uses less registers than `hipGridDim_x * hipBlockDim_x` due to a [specific optimization with the former](https://github.com/ROCm-Developer-Tools/hipamd/blob/05e3016405bdaec0b8acfe8378df968dd02136d5/include/hip/amd_detail/amd_hip_runtime.h#L337-L342).

An alternative is to pass `-mllvm -amdgpu-disable-unclustred-high-rp-reschedule` to the compiler when compiling the test, but I dont think that is a proper solution either.

Regardless, there is nothing in rocRAND related to that change to the scheduler that can really change its outcome, so the alternative is to just leave the regression until the next version of ROCm when the fix is included in the bundled version of LLVM.

@stanleytsang-amd do you have an opinion?

stanleytsang-amd (approved):
Workaround is acceptable to me for now. Approved

PR #352, opened 2023-06-13
Draft hotfix to resolve performance drop in sobol benchmarks

Naraenda:
This is a proof-of-concept band-aid to fix the performance regression seen with the new compiler by using `amdgpu_waves_per_eu`. I'm not happy with this solution but it does fix the problem.

I've mirrored @ex-rzr (Anton)'s solution to [StreamHPC/rocRAND/benchmark-performance-drop-descrete](https://github.com/StreamHPC/rocRAND/tree/benchmark-performance-drop-descrete), that resolves it without `amdgpu_waves_per_eu`, but also causes regressions in other benchmarks.


Naraenda:
Replaced with #362

PR #370, opened 2023-07-25
Use rocm-docs-core for linting

saadrahim:


saadrahim:
@MathiasMagnus I tried using the scripts but there are some issues here.

Maetveis:
I would prefer to wait for https://github.com/RadeonOpenCompute/rocm-docs-core/issues/489 instead, my reasoning being that GitHub is not the primary platform where development happens for rocRAND, so any issues that this linting would uncover would only come up during the upstreaming pull requests.

stanleytsang-amd:
Seems like there's a bunch of failures in CI and also in the files changed section.

stanleytsang-amd:
PR is stale. Closing.

PR #372, opened 2023-07-29
sobol64_direction_vector_generator.cpp: include stdint.h for uint*_t

heroxbd:
Reference: https://gcc.gnu.org/gcc-13/porting_to.html


heroxbd:
@nolmoonen  it is a trivial fix for GCC 13.  Please help review, thanks!

nolmoonen (approved):
LGTM

PR #388, opened 2023-09-15
Fix CppCheck Issues

RobsonRLemos:
This PR fixes some Cppcheck reported issues:

. Modified some functions to be static because they only access static member and do not need access a class instance.
. Added some data members to the constructor initializer list.
. Removed some warnings for exceptThrowInDestructor since the destructor is using noexcept(false).
. Removed some warnings for uninitMemberVar because some initializations are not supported.


umfranzw (approved):
This looks good to me.

stanleytsang-amd (approved):

PR #399, opened 2023-10-12
Asan Setup and Settings Map

thansen-amd:
Adding address-sanitizer builds for rocRAND

Refactoring compile code to use a settings map instead of increasingly more bool flags


jayhawk-commits:

PR #419, opened 2024-01-03
Change output type for Normal example from Int to Float

AbhinavPraveen:
The normal distribution does not output ints; set output type to float instead.

### DIFF
--- a/library/include/rocrand/rocrand.hpp
+++ b/library/include/rocrand/rocrand.hpp
@@ -3733,8 +3733,8 @@ typedef xorwow default_random_engine;
 /// int main()
 /// {
 ///     const size_t size = 8192;
-///     unsigned int * output;
-///     hipMalloc(&output, size * sizeof(unsigned int));
+///     float * output;
+///     hipMalloc(&output, size * sizeof(float));
 ///
 ///     rocrand_cpp::random_device rd;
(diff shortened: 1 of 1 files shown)


AbhinavPraveen:
Compiling the example as is produces:
```
rand2.cpp:7:5: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
    7 |     hipMalloc(&output, size * sizeof(unsigned int));
      |     ^~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
rand2.cpp:12:5: error: no matching function for call to object of type 'rocrand_cpp::normal_distribution<float>'
   12 |     dist(engine, output, size);
      |     ^~~~
/usr/include/rocrand/rocrand.hpp:489:10: note: candidate function template not viable: no known conversion from 'unsigned int *' to 'float *' for 2nd argument
  489 |     void operator()(Generator& g, RealType * output, size_t size)
      |          ^                        ~~~~~~~~~~~~~~~~~
1 warning and 1 error generated when compiling for gfx1100.
```

stanleytsang-amd:
looks good to me; paging @Naraenda to take a look

stanleytsang-amd (approved):

PR #432, opened 2024-01-23
Workaround for regressions in XORWOW and LFSR on MI200

ex-rzr:
Workaround for regressions in XORWOW and LFSR on MI200


RobsonRLemos (approved):
Tested on MI200 and the performance results look good to me.

stanleytsang-amd (approved):

PR #434, opened 2024-01-31
Switch from using hipcc to using Clang compiler

lawruble13:
As hipcc is a thin wrapper around clang, we're trying to move away from using it to reduce complexity.

### DIFF
--- a/toolchain-linux.cmake
+++ b/toolchain-linux.cmake
@@ -12,6 +12,5 @@ endif()
 
 # set(CMAKE_CXX_COMPILER "hipcc")
 # set(CMAKE_C_COMPILER "hipcc")
-set(CMAKE_CXX_COMPILER "${rocm_bin}/hipcc")
-set(CMAKE_C_COMPILER "${rocm_bin}/hipcc")
-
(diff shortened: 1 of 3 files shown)


toolchain-linux.cmake
```diff
@@ -5,13 +5,12 @@
...
+  set(llvm_bin "/opt/rocm/llvm/bin")
 endif()
 
 
-# set(CMAKE_CXX_COMPILER "hipcc")
-# set(CMAKE_C_COMPILER "hipcc")
-set(CMAKE_CXX_COMPILER "${rocm_bin}/hipcc")
-set(CMAKE_C_COMPILER "${rocm_bin}/hipcc")
-
+set(CMAKE_CXX_COMPILER "${llvm_bin}/clang++")
```
TorreZuk: Does env CXX override this?   As discussed at meeting can users easily fallback to hipcc if there are issues in custom build.

lawruble13: This is an excellent point, no this is _not_ overridden by CXX. Fixing this momentarily.

RobsonRLemos (approved):

umfranzw (approved):

PR #439, opened 2024-02-06
Updates and test coverage for hipGraph support in rocRAND

umfranzw:
Updates and test coverage for hipGraph support in rocRAND

This change allows generators to be used inside of hipGraphs.
It makes these modifications:
     * Changed all synchronous HIP API calls (mostly hipMalloc) to asynchronous ones (eg. hipMallocAsync)
     * Added an extra stream parameter to a number of functions and constructors so that the stream they issue API calls
       into can be set. The new parameter defaults to the default stream (0) so that existing callers do not need to be modified.
     * Updated unit tests to exercise the generators within hipGraphs.
    
Note: currently, the poisson distribution does not work inside of hipGraphs. This will be addressed in a future commit.
For now, I've added a note the changelog that indicates this exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,14 @@ Documentation for rocRAND is available at
 
 ### Changes
 
+* For device-side generators, you can now wrap calls to rocrand_generate_* inside of a hipGraph. There are a few 
+  things to be aware of:
+  - Generator creation (rocrand_create_generator), initialization (rocrand_initialize_generator), and destruction (rocrand_destroy_generator) must still happen outside the hipGraph.
+  - After the generator is created, you may call API functions to set its seed, offset, and order.
+  - After the generator is initialized (but before stream capture or manual graph creation begins), use rocrand_set_stream to set the stream the generator will use within the graph.
+  - A generator's seed, offset, and stream may not be changed from within the hipGraph. Attempting to do so may result in unpredicable behaviour.
+  - API calls for the poisson distribution (eg. rocrand_generate_poisson) are not yet supported inside of hipGraphs.
+  - For sample usage, see the unit tests in test/test_rocrand_hipgraphs.cpp
 * Building rocRAND now requires a C++17 capable compiler, as the internal library sources now require it. However consuming rocRAND is still possible from C++11 as public headers don't make use of the new features.
 * Building rocRAND should be faster on machines with multiple CPU cores as the library has been
   split to multiple compilation units.
--- a/test/test_rocrand_hipgraphs.cpp
+++ b/test/test_rocrand_hipgraphs.cpp
@@ -1,0 +1,145 @@
+#include <stdio.h>
+#include <gtest/gtest.h>
+
+#include <hip/hip_runtime.h>
+#include <rocrand/rocrand.h>
+
+#include "test_common.hpp"
+#include "test_rocrand_common.hpp"
+#include "test_utils_hipgraphs.hpp"
+
+class rocrand_hipgraph_generate_tests : public ::testing::TestWithParam<rocrand_rng_type> {};
+
+void test_float(std::function<rocrand_status(rocrand_generator, float*, size_t, float, float)> generate_fn, rocrand_rng_type rng_type)
+{
+    rocrand_generator generator;
+    ROCRAND_CHECK(
+        rocrand_create_generator(
+            &generator,
+            rng_type
+        )
+    );
+
+    ROCRAND_CHECK(rocrand_initialize_generator(generator));
+
+    const size_t size = 12563;
+    float mean = 5.0f;
+    float stddev = 2.0f;
+    float * data;
+    HIP_CHECK(hipMallocHelper(&data, size * sizeof(float)));
+    HIP_CHECK(hipDeviceSynchronize());
+
+    // Default stream does not support hipGraph stream capture, so create a non-blocking one
+    hipStream_t stream = 0;
+    HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
+    rocrand_set_stream(generator, stream);
+
+    hipGraphExec_t graph_instance;
+    hipGraph_t graph = test_utils::createGraphHelper(stream);
+
+    // Any sizes
+    ROCRAND_CHECK(
+        generate_fn(generator, data, 1, mean, stddev)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+    test_utils::resetGraphHelper(graph, graph_instance, stream);
+
+    // Any alignment
+    ROCRAND_CHECK(
+        generate_fn(generator, data+1, 2, mean, stddev)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+    test_utils::resetGraphHelper(graph, graph_instance, stream);
+
+    ROCRAND_CHECK(
+        generate_fn(generator, data, size, mean, stddev)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+
+    HIP_CHECK(hipFree(data));
+    ROCRAND_CHECK(rocrand_destroy_generator(generator));
+    test_utils::cleanupGraphHelper(graph, graph_instance);
+    HIP_CHECK(hipStreamDestroy(stream));
+}
+
+TEST_P(rocrand_hipgraph_generate_tests, normal_float_test)
+{
+    auto generator_fcn = [](rocrand_generator generator, float* output_data, size_t n, float mean, float stddev)
+    {
+        return rocrand_generate_normal(generator, output_data, n, mean, stddev);
+    };
+
+    test_float(generator_fcn, GetParam());
+}
+
+TEST_P(rocrand_hipgraph_generate_tests, log_normal_float_test)
+{
+    auto generator_fcn = [](rocrand_generator generator, float* output_data, size_t n, float mean, float stddev)
+    {
+        return rocrand_generate_log_normal(generator, output_data, n, mean, stddev);
+    };
+
+    test_float(generator_fcn, GetParam());
+}
+
+TEST_P(rocrand_hipgraph_generate_tests, uniform_float_test)
+{
+    const rocrand_rng_type rng_type = GetParam();
+
+    rocrand_generator generator;
+    ROCRAND_CHECK(
+        rocrand_create_generator(
+            &generator,
+            rng_type
+        )
+    );
+
+    ROCRAND_CHECK(rocrand_initialize_generator(generator));
+
+    const size_t size = 12563;
+    float * data;
+    HIP_CHECK(hipMallocHelper(&data, size * sizeof(float)));
+    HIP_CHECK(hipDeviceSynchronize());
+
+    // Default stream does not support hipGraph stream capture, so create a non-blocking one
+    hipStream_t stream = 0;
+    HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
+    rocrand_set_stream(generator, stream);
+    
+    hipGraphExec_t graph_instance;
+    hipGraph_t graph = test_utils::createGraphHelper(stream);
+
+    // Any sizes
+    ROCRAND_CHECK(
+        rocrand_generate_uniform(generator, data, 1)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+    test_utils::resetGraphHelper(graph, graph_instance, stream);
+
+    // Any alignment
+    ROCRAND_CHECK(
+        rocrand_generate_uniform(generator, data+1, 2)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+    test_utils::resetGraphHelper(graph, graph_instance, stream);
+
+    ROCRAND_CHECK(
+        rocrand_generate_uniform(generator, data, size)
+    );
+
+    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+
+    HIP_CHECK(hipFree(data));
+    ROCRAND_CHECK(rocrand_destroy_generator(generator));
+    test_utils::cleanupGraphHelper(graph, graph_instance);
+    HIP_CHECK(hipStreamDestroy(stream));
+}
+
+INSTANTIATE_TEST_SUITE_P(rocrand_hipgraph_generate_tests,
+                        rocrand_hipgraph_generate_tests,
+                        ::testing::ValuesIn(rng_types));
--- a/test/test_utils_hipgraphs.hpp
+++ b/test/test_utils_hipgraphs.hpp
@@ -1,0 +1,91 @@
+// Copyright (c) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#ifndef ROCRAND_TEST_UTILS_HIPGRAPHS_HPP
+#define ROCRAND_TEST_UTILS_HIPGRAPHS_HPP
+
+#include <hip/hip_runtime.h>
+#include "test_common.hpp"
+
+// Helper functions for testing with hipGraph stream capture.
+// Note: graphs will not work on the default stream.
+namespace test_utils
+{
+    
+    inline hipGraph_t createGraphHelper(hipStream_t& stream, const bool beginCapture=true)
+    {
+        // Create a new graph
+        hipGraph_t graph;
+        HIP_CHECK_NON_VOID(hipGraphCreate(&graph, 0));
+
+        // Optionally begin stream capture
+        if (beginCapture)
+            HIP_CHECK_NON_VOID(hipStreamBeginCapture(stream, hipStreamCaptureModeGlobal));
+
+        return graph;
+    }
+
+    inline void cleanupGraphHelper(hipGraph_t& graph, hipGraphExec_t& instance)
+    {
+        HIP_CHECK_NON_VOID(hipGraphDestroy(graph));
+        HIP_CHECK_NON_VOID(hipGraphExecDestroy(instance));
+    }
+
+    inline void resetGraphHelper(hipGraph_t& graph, hipGraphExec_t& instance, hipStream_t& stream, const bool beginCapture=true)
+    {
+        // Destroy the old graph and instance
+        cleanupGraphHelper(graph, instance);
+
+        // Create a new graph and optionally begin capture
+        graph = createGraphHelper(stream, beginCapture);
+    }
+
+    inline hipGraphExec_t endCaptureGraphHelper(hipGraph_t& graph, hipStream_t& stream, const bool launchGraph=false, const bool sync=false)
+    {
+        // End the capture
+        HIP_CHECK_NON_VOID(hipStreamEndCapture(stream, &graph));
+
+        // Instantiate the graph
+        hipGraphExec_t instance;
+        HIP_CHECK_NON_VOID(hipGraphInstantiate(&instance, graph, nullptr, nullptr, 0));
+
+        // Optionally launch the graph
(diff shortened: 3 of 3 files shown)


umfranzw:
I'm currently looking into the test failures on gfx1030, but if you'd like to have a look and leave some comments in the meantime, please feel free.

umfranzw:
I've solved the gfx1030 issues, and this change is now ready for review.

umfranzw:
I believe I've finally ironed out all the issues on gfx1030 now, but we'll have to wait and see if the checks pass.

Naraenda:
This should probably be merged after https://github.com/ROCm/rocRAND/pull/467 since the PRs might conflict. I'll poke around for some reviews too.

library/src/rocrand.cpp
```diff
@@ -433,7 +433,7 @@ rocrand_status ROCRANDAPI rocrand_get_version(int* version)
 }
 
 rocrand_status ROCRANDAPI rocrand_create_poisson_distribution(
-    double lambda, rocrand_discrete_distribution* discrete_distribution)
```
nolmoonen: This is a breaking change on a public function, probably this commented `= 0` should be uncommented?

library/src/rng/generator_type.hpp
```diff
@@ -271,7 +271,7 @@ struct rocrand_generator_impl_base
         return m_stream;
     }
 
-    void set_stream(hipStream_t stream)
+    virtual void set_stream(hipStream_t stream)
```
nolmoonen: Why should this be `virtual`, which derived class overrides it?

library/src/rng/distribution/discrete.hpp
```diff
@@ -176,20 +177,20 @@ class rocrand_discrete_distribution_base : public rocrand_discrete_distribution_
             hipError_t error;
             if ((Method & ROCRAND_DISCRETE_METHOD_ALIAS) != 0)
             {
-                error = hipMalloc(&probability, sizeof(double) * size);
+                error = hipMallocAsync(&probability, sizeof(double) * size, m_stream);
```
nolmoonen: On NVIDIA platforms, this fails with
```
[build] ../../library/src/rng/distribution/discrete.hpp(180): error: argument of type "double **" is incompatible with parameter of type "void **"
[build]                   error = hipMallocAsync(&probability, sizeof(double) * size, m_stream);
[build]                                          ^
[build] 
```
As far as I can tell, the templatized `hipMallocAsync` is not available on NVIDIA platforms. https://github.com/ROCm/HIP/blob/11d368b6ac20553d3748f6492b96d19738fc78a5/include/hip/hip_runtime_api.h#L9090 is guarded with `!defined(__HIP_PLATFORM_NVIDIA__)` and https://github.com/ROCm/hipother/blob/develop/hipnv/include/hip/nvidia_detail/nvidia_hip_runtime_api.h only contains the `void**` function. This seems to be an error in HIP, but can you workaround this with a `reinterpret_cast<void**>`?

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ Documentation for rocRAND is available at
 
 ### Changes
 
+* Generators are now capable of operating inside of non-blocking streams. You can either passing a stream to the constructor or use the set_stream member function.
```
nolmoonen: These changes cause our (internal) HIP Windows CI to fail, I've not yet had a chance to have a closer look. Are Windows tests ran on this GitHub PR and do they pass? (I cannot view pipeline details.)

library/src/rocrand.cpp
```diff
@@ -480,7 +481,8 @@ rocrand_status ROCRANDAPI
     rocrand_create_discrete_distribution(const double*                  probabilities,
                                          unsigned int                   size,
                                          unsigned int                   offset,
-                                         rocrand_discrete_distribution* discrete_distribution)
+                                         rocrand_discrete_distribution* discrete_distribution,
+                                         hipStream_t                    stream /* = 0 */)
```
nolmoonen: Same here.

library/src/rocrand.cpp
```diff
@@ -526,7 +529,7 @@ rocrand_status ROCRANDAPI
 }
 
 rocrand_status ROCRANDAPI
-    rocrand_destroy_discrete_distribution(rocrand_discrete_distribution discrete_distribution)
+    rocrand_destroy_discrete_distribution(rocrand_discrete_distribution discrete_distribution, hipStream_t stream /* = 0 */)
```
nolmoonen: And here.

test/internal/test_rocrand_lfsr113_prng.cpp
```diff
@@ -30,13 +30,16 @@
 
 #include "test_common.hpp"
 #include "test_rocrand_common.hpp"
+#include "test_utils_hipgraphs.hpp"
 
-struct rocrand_lfsr113_prng_tests : public testing::TestWithParam<rocrand_ordering>
+#include <iostream>
+
+struct rocrand_lfsr113_prng_tests : public testing::TestWithParam<std::tuple<rocrand_ordering, bool>> // TODO add a bool param here
```
nolmoonen: TODO is completed?

test/test_rocrand_kernel_mtgp32.cpp
```diff
@@ -208,13 +212,26 @@ TEST(rocrand_kernel_mtgp32, rocrand)
     HIP_CHECK(hipMallocHelper(&output, output_size * sizeof(unsigned int)));
     HIP_CHECK(hipDeviceSynchronize());
 
+    hipStream_t stream = 0;
```
nolmoonen: I think it would be good to also perform `rocrand_make_state_mtgp32` inside the graph, now that it has an option to pass the stream.

library/src/rocrand.cpp
```diff
@@ -480,7 +481,8 @@ rocrand_status ROCRANDAPI
     rocrand_create_discrete_distribution(const double*                  probabilities,
                                          unsigned int                   size,
                                          unsigned int                   offset,
-                                         rocrand_discrete_distribution* discrete_distribution)
+                                         rocrand_discrete_distribution* discrete_distribution,
+                                         hipStream_t                    stream /* = 0 */)
 {
     if(discrete_distribution == NULL)
     {
```
nolmoonen: Shouldn't the stream be passed to `rocrand_discrete_distribution_base`'s constructor?

library/src/rocrand.cpp
```diff
@@ -536,10 +539,11 @@ rocrand_status ROCRANDAPI
...
-                      sizeof(rocrand_discrete_distribution_st),
-                      hipMemcpyDefault);
+    error = hipMemcpyAsync(&h_dis,
+                           discrete_distribution,
+                           sizeof(rocrand_discrete_distribution_st),
+                           hipMemcpyDefault,
+                           stream);
     if(error != hipSuccess)
     {
         return ROCRAND_STATUS_INTERNAL_ERROR;
```
nolmoonen: Shouldn't the stream be set on `h_dis` before calling `h_dis.deallocate();`?

library/src/rng/distribution/discrete.hpp
```diff
@@ -100,15 +101,15 @@ class rocrand_discrete_distribution_base : public rocrand_discrete_distribution_
         {
             if (probability != NULL)
             {
-                ROCRAND_HIP_FATAL_ASSERT(hipFree(probability));
+                ROCRAND_HIP_FATAL_ASSERT(hipFreeAsync(probability, m_stream));
```
nolmoonen: I think it would be good to make these malloc and free calls asynchronous, and address the Poisson issue at the same time.

With these changes, the stream is not set for the `poisson_distribution_manager` in the derived classes from `rocrand_generator_impl_base` (for example `rocrand_lfsr113_template`).

If `generate_poisson` is called and later the destructor of `rocrand_lfsr113_template`, the members of `rocrand_poisson_distribution` are freed on the default stream. Based on your comments on the hipCUB PR, I would assume that`hipFreeAsync(defaultStream)` will wait on a kernel dispatched on a non-blocking stream.

I think it's much better to make this explicit. Since you have excluded this from the tests, was there a specific case that fails?

library/src/rocrand.cpp
umfranzw: The default value of 0 for this parameter is actually defined in library/include/rocrand/rocrand.h (which this file includes). Unfortunately, uncommenting the "= 0" here results in a multiple-definition compiler error. I reason I included the commented-out version here in the cpp file was just to emphasize that it exists in the header.

library/src/rng/generator_type.hpp
umfranzw: Nice catch, thanks. This was left over from a previous version of the change, sorry about that. I'll push a follow-up commit shortly that fixes up some of the issues with the poisson distribution, which make it necessary again. When the user calls set_stream() on a subclass of rocrand_generator_impl_base (eg. lfsr113_prng), we need to let the subclass' m_poisson data member know that the stream has changed. We can do that by overriding this function so that it calls m_poisson.set_stream(new_stream).

library/src/rng/distribution/discrete.hpp
umfranzw: Thanks for the heads up about this. In the new commit I'll push shortly, I've updated all occurrences of hipMallocAsync so they use the cast you suggested.

CHANGELOG.md
umfranzw: Yes, the Windows tests are run. The tests failed, but the test systems have been flakey lately. I'll take another look at them after I push the next commit (which should re-trigger them) to address the other issues mentioned here.

library/src/rocrand.cpp
umfranzw: The default value for this one is also defined in the rocrand.h header file, as described above.

test/internal/test_rocrand_lfsr113_prng.cpp
umfranzw: I was sure I did a search for all occurrences of this... my apologies. I've removed it in the next commit I'll push shortly.

test/test_rocrand_kernel_mtgp32.cpp
umfranzw: Nice catch, thanks. I've moved this inside the stream capture block so it's run inside the graph.

library/src/rocrand.cpp
umfranzw: Fixed, thanks.

library/src/rocrand.cpp
umfranzw: Fixed, thank you.

library/src/rng/distribution/discrete.hpp
umfranzw: I've managed to fix up some of the issues with the Poisson distribution, such the all tests in test/internal/test_poisson_distribution.cpp now pass with graphs. I'll push a new commit shortly.
I'll also leave in the failing test case for lfsr113_prng (the "test_poisson" unit test) so that you can see it at runtime. If you have any insight into what might be causing it, I'd really appreciate any feedback you have.

It's not just lfsr113_prng that has this problem - it also happens in the same test case for other generators. I've played around with the placement of the stream capture block (capturing progressively narrower slices of the generator code), but I still haven't been able to nail down exactly what the problem is.

library/src/rocrand.cpp
nolmoonen: Ah I see, my bad, I didn't realize the function declaration is elsewhere.

library/src/rng/distribution/discrete.hpp
nolmoonen: Looking further into this, the problem is that the host data used in the `hipMemcpyAsync` in `create_alias_table` (and `create_cdf`) is already destroyed by the time the graph is launched. This should also cause a fail in the `TEST_P(rocrand_generate_poisson_tests, generate_test)` test. In my testing it does not fail, I'm not sure why, but the host memo

I'm not sure if this can be solved in a satisfying way. I think the best thing to do would be to make an exception for `rocrand_generate_poisson` after all. We could later extend the API to add an `rocrand_generate_discrete` that uses the `rocrand_discrete_distribution` constructed by `rocrand_create_poisson_distribution` or `rocrand_create_discrete_distribution`. It should then be safe to record and launch the `rocrand_generate_discrete` calls.

nolmoonen: Similar problems exist in e.g. `TEST_P(rocrand_kernel_xorwow_poisson, rocrand_discrete)`, where `rocrand_destroy_discrete_distribution` is captured in the stream, but this will destroy the host input for the copy. I think the test should look like:
```
rocrand_create_poisson_distribution

createGraphHelper
rocrand_discrete_kernel
endCaptureGraphHelper

rocrand_destroy_discrete_distribution

copy h2d and test
```

test/test_rocrand_kernel_sobol32.cpp
```diff
@@ -318,13 +382,26 @@ TEST(rocrand_kernel_sobol32, rocrand_log_normal)
...
+        HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
+        graph = test_utils::createGraphHelper(stream);
+    }
+
     hipLaunchKernelGGL(
         HIP_KERNEL_NAME(rocrand_log_normal_kernel<state_type>),
         dim3(8), dim3(32), 0, 0,
         output, m_vector, output_size
     );
     HIP_CHECK(hipGetLastError());
```
Naraenda: This kernel launch should be synced with `stream` right?
```suggestion
    hipLaunchKernelGGL(
        HIP_KERNEL_NAME(rocrand_log_normal_kernel<state_type>),
        dim3(8), dim3(32), 0, stream,
        output, m_vector, output_size
    );
    HIP_CHECK(hipGetLastError());
```

test/test_rocrand_kernel_sobol64.cpp
```diff
@@ -391,13 +481,28 @@ TEST_P(rocrand_kernel_sobol64_poisson, rocrand_poisson)
...
+        HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
+        graph = test_utils::createGraphHelper(stream);
+    }
+
     hipLaunchKernelGGL(
         HIP_KERNEL_NAME(rocrand_poisson_kernel<state_type>),
         dim3(8), dim3(32), 0, 0,
         output, m_vector, output_size, lambda
     );
     HIP_CHECK(hipGetLastError());
```
Naraenda: Idem.
```suggestion
    hipLaunchKernelGGL(
        HIP_KERNEL_NAME(rocrand_poisson_kernel<state_type>),
        dim3(8), dim3(32), 0, stream,
        output, m_vector, output_size, lambda
    );
    HIP_CHECK(hipGetLastError());
```

nolmoonen (commented):
After some internal discussion with the team on how hipGraph should be handled in rocRAND, some more fundamental points came up, that I'd like to discuss.

The first point is that, from a user's perspective, there is little to be gained from capturing the generator creation and destruction in a graph for multiple `rocrand_generate` calls. In fact, if the creation or destruction is captured (or any other functions implicitly causing re-initialization), a hidden overhead is added. It makes more sense to force initialization and keep the generator state alive. Is there a specific need to support hipGraph for all device functionality? If not, it might be better to drop support for create/destroy generator (effectively dropping support for the `init` calls) and create/destroy for distributions. It would simplify the PR and usage lot. Then, the complete support for rocRAND would look like:

```c++
rocrand_create_generator(g, ...);
rocrand_set_seed(g, ...)
rocrand_set_offset(g, ...)
rocrand_set_ordering(g, ...)
// initialization must be done before capturing
rocrand_initialize_generator(g); 

hipStreamCreate(&stream);
rocrand_set_stream(g, stream);

hipStreamBeginCapture(stream);
// stream should not be changed
rocrand_generate(g, output, size); // as many generate calls
// rocrand_generate_poisson remains an exception
// rocrand_set_seed, rocrand_set_offset, and rocrand_set_ordering are not allowed, see below
hipStreamEndCapture(stream);

// replay the capture as many times as you want
hipStreamSynchronize(stream); // wait for the replay(s) to finish

// destroy state only after replay is finished
rocrand_destroy_generator(g);
```

It is possible to do a refactor such that `init` can be supported without doing a device malloc/free, which would enable setting seed, offset, and ordering for hipGraph without overhead. But this is something we can add later. It would also add support for the C++ wrappers, which currently have no way to force initialization.

The second point is that adding graph checks to every test adds a lot of noise to the implementation files, and increases the running time of the tests. It should be sufficient to only check the public API, dropping the checks in `test/internal`. Furthermore, the kernel tests (for the "kernel API") use no HIP API functionality outside of launching kernels, so the hipGraph checks there are overkill. Finally, each generator uses the same HIP API calls for the different distribution/data types (exception being Poisson, which is not supported for the time being). The best thing might be to make a standalone test for hipGraph, that tests the public API for each generator with a single distribution type.

I realize it would have been more productive to add these notes upon first review (or even in a discussion beforehand), I am sorry for that.

Maetveis:
> It is possible to do a refactor such that `init` can be supported without doing a device malloc/free

Correction: The first call to init() would still have to allocate, only that changing seed, or offset wouldn't need to.

Ordering is also special because that can influence the amount of state we need so it also could require re-allocation.

> It would also add support for the C++ wrappers, which currently have no way to force initialization.

Due to point above it wouldn't.

library/src/rng/distribution/discrete.hpp
umfranzw: Thanks @nolmoonen, I appreciate your help with tracking this problem down. I spoke with Colin a while back, and he'd mentioned it should be fine to make an exception for the poisson distribution like you suggested, just as long as we add a note about it to the changelog. Things are quite busy right now, but I'll do my best to have that up shortly.

test/test_rocrand_kernel_sobol32.cpp
umfranzw: Nice catch, thank you. This must have gotten lost when I rebased. Given @nolmoonen's point below about hipGraph testing in the kernel test files being a little overkill, I will remove all hipGraph tests from this file.

test/test_rocrand_kernel_sobol64.cpp
umfranzw: Thanks. As above - I'll remove all hipGraph tests from the "kernel test" files.

umfranzw:
@nolmoonen Those are really good points, thank you. Here are a few thoughts in response:

> It makes more sense to force initialization and keep the generator state alive. Is there a specific need to support hipGraph for all device functionality? ... it might be better to drop support for create/destroy generator (effectively dropping support for the init calls) and create/destroy for distributions. It would simplify the PR and usage lot.

I completely agree - I was a little uncertain if moving initialization inside hipGraphs was necessary, and I decided to try to go for it, but in retrospect it was much more trouble that it was worth. I'm pretty certain that we don't need to support hipGraph for _all_ device functionality right now - as long as we note the exceptions in the changelog so that users are aware. I'll update the PR to reflect your suggestions.

> ... adding graph checks to every test adds a lot of noise to the implementation files, and increases the running time of the tests.

I agree with the points you made about this as well. I'll remove the redundant tests, leaving only coverage for the public API, in the form of a standalone hipGraph test, that exercises the public API for each generator with a single distribution type, as you've suggested.

> I realize it would have been more productive to add these notes upon first review (or even in a discussion beforehand), I am sorry for that.

No problem - I should have run some of these questions by you at the points I made of the decisions we've been discussing. This is my first real contribution to the repo, so at the outset I didn't always have quite a large enough picture of the API in my head to make a fully informed decision. In the future, I'll make sure that when this is the case, I check in with you folks more before choosing a fork in the road.

umfranzw:
I've completed the requested changes.

test/test_rocrand_hipgraphs.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    HIP_CHECK(hipFree(data));
+    ROCRAND_CHECK(rocrand_destroy_generator(generator));
+    test_utils::cleanupGraphHelper(graph, graph_instance);
+    HIP_CHECK(hipStreamDestroy(stream));
+}
+
+INSTANTIATE_TEST_SUITE_P(rocrand_hipgraph_generate_tests,
+                        rocrand_hipgraph_generate_tests,
+                        ::testing::ValuesIn(rng_types));
```
stanleytsang-amd: missing newline at end of file?

CHANGELOG.md
```diff
@@ -44,6 +44,14 @@ Documentation for rocRAND is available at
 
 ### Changes
 
+* For device-side generators, you can now wrap calls to rocrand_generate_* inside of a hipGraph. There are a few
```
stanleytsang-amd, line 47: I think we should consider adding an official documentation page in /docs on using hipGraphs with rocRAND (and also for the other Stream libraries, for that matter).  Essentially with the notes in the changelog entry. The information in the changelog is very useful, but would be very hard to find by an end-user in the future.

The deadline for 6.2 is getting pretty close so maybe we can do this for 6.3.

stanleytsang-amd (approved):
LGTM, if @nolmoonen can give one final look at this before EOD Monday that would be great, thanks!

CHANGELOG.md
nolmoonen: I'll make sure we internally track documenting this (for rocPRIM as well, which AFAICT is the only other library we have been adding graph support for so far) and adding support for generating with a Poisson distribution.

nolmoonen (approved):
LGTM

CHANGELOG.md
```diff
@@ -44,6 +44,14 @@ Documentation for rocRAND is available at
 
 ### Changes
 
+* For device-side generators, you can now wrap calls to rocrand_generate_* inside of a hipGraph. There are a few 
+  things to be aware of:
+  - Generator creation (rocrand_create_generator), initialization (rocrand_initialize_generator), and destruction (rocrand_destroy_generator) must still happen outside the hipGraph.
```
Naraenda, line 49: Might be useful to detect stream captures with `hipStreamGetCaptureInfo_v2` and throw an appropriate error when it happens. But this is non blocking IMHO.

Naraenda (approved):
Small comment, but can be deferred to 6.3. Otherwise LGTM

test/test_rocrand_hipgraphs.cpp
umfranzw: Fixed, thanks.

CHANGELOG.md
umfranzw: Thanks @Naraenda, that's a really good idea. Given how close we are to the deadline for 6.2, maybe I'll defer this to 6.3 if that's ok. I'll make a note of it.

umfranzw: Thanks @Naraenda, that's a really good idea. Given how close we are to the deadline for 6.2, maybe I'll defer this to 6.3 if that's ok. I'll make a note of it.

PR #443, opened 2024-02-09
Enable compatibility with projects using HIP language

lawruble13:
Previously if a project tried to include rocRAND via `add_subdirectory`, but was using the `HIP` language provided by CMake, this would fail (see [rocThrust issue #238](https://github.com/ROCmSoftwarePlatform/rocThrust/issues/238)). This PR should enable rocRAND to be used in this fashion.


CMakeLists.txt
```diff
@@ -80,9 +99,9 @@ include(CMakeDependentOption)
 include(cmake/Dependencies.cmake)
 
 # Detect compiler support for target ID
-if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
-    execute_process(COMMAND ${CMAKE_CXX_COMPILER} "--help"
-    OUTPUT_VARIABLE CXX_OUTPUT
+if( CMAKE_${ROCRAND_LANG}_COMPILER MATCHES ".*/hipcc$" )
```
cgmb, line 102: hipcc cannot be used as the HIP language compiler.

CMakeLists.txt
```diff
@@ -57,10 +57,29 @@ list(APPEND CMAKE_MODULE_PATH
     $ENV{ROCM_PATH}/llvm
 )
 
+# Detect if already using HIP language
+get_property(ENABLED_LANGUAGES GLOBAL PROPERTY ENABLED_LANGUAGES)
+if(HIP IN_LIST ENABLED_LANGUAGES)
```
cgmb, line 62: Could we maybe just finish up #266 and suggest that users to set `-DROCRAND_CMAKE_HIP_LANGUAGE=ON`?

haampie:
It would be good to fix this or #266.

In the [Spack package manager](https://github.com/spack/spack) we're improving hip support by modeling it as a language, but I'm somewhat surprised that 4 years after CMake started support hip as a language, AMD's own packages do not use it :thinking:

jayhawk-commits:

lawruble13:
Ditto Cory's comments on #266, this is still a desirable feature but this PR is very out of date and should probably be redone in its entirety.

PR #467, opened 2024-03-22
Upstream staging 2024 03 22

parbenc:
**Notable changes**
* Added `rocrand_create_generator_host_blocking` which dispatches without stream semantics.
* Added host-side generator for `ROCRAND_RNG_PSEUDO_MTGP32`.
* Added offset and skipahead functionality to LFSR113 generator.
* Added dynamic ordering for architecture `gfx1102`.
* Added the following functions for the `ROCRAND_RNG_PSEUDO_MTGP32` generator:
  * `rocrand_normal2`
  * `rocrand_normal_double2`
  * `rocrand_log_normal2`
  * `rocrand_log_normal_double2`

* Rename and unify the existing ROCRAND_DETAIL_.*_BM_NOT_IN_STATE to ROCRAND_DETAIL_BM_NOT_IN_STATE
* Deprecated internal header: src/rng/distribution/distributions.hpp
* Deprecated internal header: src/rng/device_engines.hpp

**Known issues**
* SOBOL64 and SCRAMBLED_SOBOL64 generate poisson-distributed `unsigned long long int` numbers instead of `unsigned int`. This will be fixed in the next major release.


library/include/rocrand/rocrand_common.h
```diff
@@ -41,10 +41,6 @@
 
 #define ROCRAND_KERNEL __global__ static
 
-#ifndef FQUALIFIERS
-#define FQUALIFIERS __forceinline__ __device__
```
stanleytsang-amd, line 43: @parbenc @Naraenda Why is FQUALIFIERS being removed?  It seems like a useful and heavily-used macro.

parbenc: The main problem is that it is defined in multiple locations, with different content. Another issue is that it is less readable, than just writing out `__forceinline__ __device__`. In the end we decided, that it is better to just remove it.

stanleytsang-amd: Okay.

stanleytsang-amd (approved):

PR #482, opened 2024-04-22
Updates and test coverage for hipGraph support in rocRAND (#439) <cherry-picked from develop>

umfranzw:
This change allows device-side generators to be used inside of hipGraphs. More specifically, you can wrap calls to rocrand_generate_* inside of a hipGraph. There are a few things to be aware of:
  - Generator creation (rocrand_create_generator), initialization (rocrand_initialize_generator), and destruction (rocrand_destroy_generator) must still happen outside the hipGraph.
  - After the generator is created, you may call API functions to set its seed, offset, and order.
  - After the generator is initialized (but before stream capture or manual graph creation begins), use rocrand_set_stream to set the stream the generator will use within the graph.
  - A generator's seed, offset, and stream may not be changed from within the hipGraph. Attempting to do so may result in unpredicable behaviour.
  - API calls for the poisson distribution (eg. rocrand_generate_poisson) are not yet supported inside of hipGraphs.

I've added a note to the changelog that mentions these details. In addition to the changes necessary to support the behaviour described above, this change also:
- updates the changelog to alert the user to the restrictions mentioned above
- adds new unit test coverage to exercises generators and distributions within hipGraphs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,14 @@ Documentation for rocRAND is available at
 
 ### Changes
 
+* For device-side generators, you can now wrap calls to rocrand_generate_* inside of a hipGraph. There are a few 
+  things to be aware of:
+  - Generator creation (rocrand_create_generator), initialization (rocrand_initialize_generator), and destruction (rocrand_destroy_generator) must still happen outside the hipGraph.
+  - After the generator is created, you may call API functions to set its seed, offset, and order.
+  - After the generator is initialized (but before stream capture or manual graph creation begins), use rocrand_set_stream to set the stream the generator will use within the graph.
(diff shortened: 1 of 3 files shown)


RobsonRLemos (approved):
LGTM.

stanleytsang-amd (approved):

stanleytsang-amd:
@randyh62 Could you review this please?  Need an approval from docs team.

CHANGELOG.md
randyh62, line 1: unpredictable

spelling

randyh62 (approved):
Just a spelling error.

CHANGELOG.md
stanleytsang-amd: @umfranzw is on vacation until May 6 and we need to get this PR in now for 6.2 - I'll make a note we should get this typo fixed in develop branch.

PR #483, opened 2024-04-22
Update device-specific config headers with benchmark tuning results

umfranzw:
I ran the benchmark tuning code and scripts on the remaining architectures in our GPU support matrix. This change updates the config headers with the results.


RobsonRLemos (approved):
LGTM.

PR #404, opened 2023-10-30
StreamHPC 2023-10-30 (Dynamic Ordering Groundwork)

Naraenda:
This PR contains the groundwork for dynamic ordering. XORWOW has been changed to use dynamic ordering. There will be a second PR that contains the bulk changes on the other generators.

--- 

This PR includes the following merges, ordered from new to old. Noteable changes are bolded.

- 73341441212acb59a6474fddea07c9f8bd2c2add Refactor `if`-chains in `rocrand.cpp`.
- 8ad8f2c8982cc10fa68a006f0e5bec01768b9cad Various doxygen fixes to resolve warnings.
- b4327f3c696cab15cecfdc9c623a87b9e9e90bc4 Fix narrowing conversion in Threefry distribution
- 35cc0305ed4495505c17ebcc02a8668b7b28adf7, b0a08235db021292a1b1298b890fbc080034e38a, 553b10b420a94fe8c290bf49d17b5e6e432b50b1 **XORWOW dynamic ordering implementation**
- 4a0dbc4af676369b50bb433994baecf52c2a170b **Config generator for dynamic ordering**
- ce68cce14826d8452fb8b37bd18cf6008495c9f8 C++ 2017
- 8ddbc852d832caed2ee571e01749ed88d9201004 **Dynamic ordering changes**

---

We still want to add a minor thing to this PR:
- [ ] Config tuning optimizations.

PR #485, opened 2024-05-03
Workaround for regression in Sobol kernel

nolmoonen:
https://github.com/ROCm/rocRAND/pull/404 introduces a host generator for the Sobol variants. Through a small refactor, the host and device generator use the same function for generation. This adds one level of indirection for the device kernel, which results in a slightly different code generation by the compiler. In turn, the performance is diminished in a specific case. This PR adds a workaround, removing this indirection for the device generator.


stanleytsang-amd (approved):
verified scrambed_sobol drops are recovered. thanks!

PR #487, opened 2024-05-03
Cherrypick sobol workaround to 6.2

stanleytsang-amd:
This regains the performance drop seen in benchmarks


RobsonRLemos (approved):
LGTM.

PR #492, opened 2024-05-14
Fixed NULL check

rponnuru5:
Added a test case to check for NULL for rocrond_generate_long_long and rocrand_geenrate_char

### DIFF
--- a/test/test_rocrand_generate.cpp
+++ b/test/test_rocrand_generate.cpp
@@ -107,6 +107,14 @@ TEST_P(rocrand_generate_tests, char_test)
 
 TEST_P(rocrand_generate_tests, short_test)
 {
+    const size_t size = 12563;
+    unsigned short * data = NULL;
+    rocrand_generator generator = NULL;
+    EXPECT_EQ(
+        rocrand_generate(generator, (unsigned short *) data, size),
+        ROCRAND_STATUS_NOT_CREATED
+    );
+
     const rocrand_rng_type rng_type = GetParam();
 
     rocrand_generator generator;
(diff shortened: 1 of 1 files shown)


test/test_rocrand_generate.cpp
```diff
@@ -69,18 +69,23 @@ TEST_P(rocrand_generate_tests, int_test)
 
 TEST_P(rocrand_generate_tests, char_test)
 {
+    const size_t size = 12563;
+    unsigned char * data = NULL;
```
Naraenda:
```suggestion
    unsigned char * data = nullptr;
```

Idem for all other uses of `NULL`.

Naraenda:
I meant replacing _all_ the `NULL` with `nullptr` in the proposed change, not just that single one 😅

rponnuru5:
Hi Narendra, I have updated in all the places. pls review and approve the PR.

test/test_rocrand_generate.cpp
```diff
@@ -145,10 +156,9 @@ TEST_P(rocrand_generate_tests, short_test)
 
 TEST(rocrand_generate_tests, neg_test)
```
Naraenda, line 157: After thinking on this a bit longer, why was it chosen to add the null check in the existing tests and not extend the "null" test into a [typed test suite](http://google.github.io/googletest/reference/testing.html#TYPED_TEST_SUITE)?

stanleytsang-amd: @rponnuru5 would you be able to refactor your changes to meet Nara's suggestion?

rponnuru5: @Naraenda , Pls provide more details about it.

Naraenda: I meant grouping the `nullptr` tests together with the already existing `nullptr` tests. Having the `nullptr` test in the generator parameterized tests doesn't make sense since we don't instantiate the different types of generators. There's no point in testing the same things multiple times with the exact same input parameters. I've implemented it here: https://github.com/ROCm/rocRAND/pull/519

rponnuru5:
It is handled as part of 519 PR so closing this

PR #496, opened 2024-05-17
Rename test_config_dispatch.cpp to test_rocrand_config_dispatch.cpp

stanleytsang-amd:
This test has the exact same name as a test in [rocPRIM](https://github.com/ROCm/rocPRIM/blob/develop/test/rocprim/test_config_dispatch.cpp)  which can lead to conflicts when installing the tests/clients packages.


lawruble13 (approved):
Approved to merge once tests pass.

RobsonRLemos (approved):
Approved when the checks are completed.

PR #497, opened 2024-05-17
Cherry-pick performance workarounds to 6.2

stanleytsang-amd:
A compiler change caused the generate function to no longer be inlined which caused a performance regression. Force inlining resolves this.


NguyenNhuDi (approved):

RobsonRLemos (approved):
Approved to merge when the checks are completed.

PR #508, opened 2024-06-21
Hotfix: TBB dependency

lawruble13:
#404 added a dependency on TBB, which is silently included or not based on whether it's present at build time, and which was not captured in the package dependencies. This fix allows for specifying whether TBB should be used in compilation at configure time, and if so TBB is added to the package dependencies.


RobsonRLemos (approved):
The hotfix for TBB dependency looks good to me.

PR #509, opened 2024-06-26
Enable gfx12 support

stanleytsang-amd:


umfranzw (approved):
This looks good to me - the only thing we changed for rocRAND was adding cmake build targets.

PR #515, opened 2024-07-03
Set default target architectures for ASAN build

umfranzw:
Device-side address sanitizer instrumentation requires xnack+. Set the default target GPU architectures to those that provide xnack.


stanleytsang-amd (approved):
lgtm, just need to fix merge conflict with develop

PR #534, opened 2024-08-09
Missing precision support page

neon60:


neon60:
So the reported problem already have been fixed:
https://github.com/ROCm/rocRAND/commit/a29658990144a63e02047ee2265f7d5adb8df8ab

Great work!

PR #536, opened 2024-08-16
reformated device_api and host_api benchmark to include engine, distribution, mode, throughput gigabytes per second, lambda columns

NguyenNhuDi:


benchmark/benchmark_rocrand_host_api.cpp
```diff
@@ -23,186 +23,187 @@
...
 #include <map>
+#include <rocrand/rocrand.h>
 #include <string>
 #include <vector>
 
 #ifndef DEFAULT_RAND_N
 const size_t DEFAULT_RAND_N = 1024 * 1024 * 128;
 #endif
 
+typedef std::unique_ptr<benchmark::BenchmarkReporter> PtrType;
```
umfranzw: Just wanted to double check - is this type still used? I wasn't able to find any other occurrences of it in this change.

benchmark/custom_csv_formater.hpp
```diff
@@ -0,0 +1,182 @@
...
+      default:
+        tmp += c;
+        break;
+      }
+    }
+    return '"' + tmp + '"';
+  }
+
+  // Function to return an string for the calculated complexity
+  std::string GetBigOString(BigO complexity) {
```
umfranzw: This is not a big deal, but I think the parameter (complexity) could be made const here, since it's not modified inside the function.

benchmark/custom_csv_formater.hpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  void PrintRunData(const Run &report);
+  bool printed_header_;
+  std::set<std::string> user_counter_names_;
+
+  std::ostream *nullLog = nullptr;
+
+  std::vector<std::string> elements = {
```
umfranzw: Just a suggestion, but since this vector doesn't change, it might be a nice to use std::array instead and make it const, eg:
```
#include <array>
const std::array<const std::string, 12> elements = { ... };
```

benchmark/custom_csv_formater.hpp
```diff
@@ -0,0 +1,182 @@
...
+  std::string temp = run.benchmark_name();
+  temp.erase(0, temp.find("<") + 1);
+  
+  std::string engineName = std::string(temp.begin(), temp.begin() + temp.find(","));
+
+  temp.erase(0, engineName.size() + 1);
+  temp.erase(0, temp.find(",") + 1);
+  std::string disName = std::string(temp.begin(), temp.begin() + temp.find(">"));
+
+
```
umfranzw: Just a nitpicky thing, but it might be nice to remove the extra newline here.

umfranzw (changes requested):

benchmark/benchmark_rocrand_host_api.cpp
NguyenNhuDi: It is not being used anymore, forgot to remove it!

benchmark/custom_csv_formater.hpp
NguyenNhuDi: done

benchmark/custom_csv_formater.hpp
NguyenNhuDi: done

benchmark/custom_csv_formater.hpp
NguyenNhuDi: done!

umfranzw (approved):

umfranzw (changes requested):
Hi Di, I think this looks great - just one quick question though. 
I noticed that I do not see the new columns show up when I run with:
./benchmark/benchmark_rocrand_device_api --benchmnark_format=csv

Is there anything special I'd need to do to show them?

NguyenNhuDi:
Hi @umfranzw, you are right, google benchmark distinguishes between --benchmark_format and --benchmark_out_format. 
--benchmark_format will change what appears on the terminal, while --benchmark_out_format will change what is outputted to your output file. I am working on changing up --benchmark_format as well and should have a new commit shortly

benchmark/custom_csv_formater.hpp
```diff
@@ -0,0 +1,181 @@
+#pragma once
```
umfranzw, line 21: I gave this another try and everything seems to work great! There's just one more little thing I noticed that I missed on the first pass yesterday (sorry) - I think this file is missing the license/copyright description at the top. You should be able to just copy/paste it from another header. Just need to make sure the copyright year is up to date (2024).

umfranzw (approved):
This looks good to me.

NguyenNhuDi:
As per @RobsonRLemos , we have decided to add 3 more columns, Mode, gigabytes per second and Lambda for consistency with legacy benchmarks. @umfranzw  could you review these changes? 
Thanks!

NguyenNhuDi:
sorry @umfranzw miss clicked

RobsonRLemos (approved):
The columns included in the benchmark output are looking good. Thanks.

benchmark/custom_csv_formater.hpp
```diff
@@ -154,16 +157,34 @@ void customCSVReporter::PrintRunData(const Run& run)
...
+    if(deviceName != "device_kernel")
+    {
+        mode = std::string(temp.begin(), temp.begin() + temp.find(','));
+        temp.erase(0, temp.find(",") + 1);
+    }
     std::string disName = std::string(temp.begin(), temp.begin() + temp.find(">"));
 
-    Out << engineName << ",";
-    Out << disName << ",";
+    std::string lambda = "";
```
umfranzw, line 178: This is probably not significant enough to require a change, but just for future reference, I believe C++ strings have a constructor that initializes them to the empty string, so you don't need to set them like this.

benchmark/custom_csv_formater.hpp
```diff
@@ -154,16 +157,34 @@ void customCSVReporter::PrintRunData(const Run& run)
...
+
+    size_t ePos = disName.find("=");
+    if(ePos <= disName.size())
+    {
+        lambda = std::string(disName.begin() + (ePos + 1), disName.end() - 1);
```
umfranzw, line 183: I think the string parsing in this function is fine for this task, because we're expecting things to always be present and always be in the same order. However, just for future reference, using a regular expression with capture groups can be really useful in situations where you need to pull out a bunch of sub-strings like this. You can create a regex pattern that contains all the delimiters ('<', ','), apply it to the original temp string, and then iterate through the captured text in a single loop.

umfranzw:
> sorry @umfranzw miss clicked

No worries, I took a look at the new changes, and I think everything still looks good.

benchmark/custom_csv_formater.hpp
NguyenNhuDi: okay thank you, sounds good

benchmark/custom_csv_formater.hpp
NguyenNhuDi: okay, thank you

PR #540, opened 2024-08-23
Clarify HIP usage

lawruble13:
The Windows pathway should no longer require special treatment, so we can simplify by moving everything into VerifyCompiler.cmake. Additionally, we shouldn't need to support the use of the HIP CMake package as a module, config should be sufficient. Also added some comments to clarify what is going on.


umfranzw (approved):
I think this looks good. I tried a test build on Windows and it seems to work fine.

PR #512, opened 2024-07-01
Develop stream 2024-07-01

Naraenda:
This PR brings various updates, intended for ROCm 6.3.

It contains the following merge commits:

* 0350e0f5 perf: improve performance on rocm 6.2.
* ade3f6b2 fix: performance regression introduced with hip graph support.
* c90e74c1 docs: add documentation for hip graph support
* fa13dd77 feat: hip graph support for poisson distribution
* 4224428f feat: host implementation for mt19937
* 615fc5a7 perf: use alias method instead of binary search in cdf
* 8f7e2af7 docs: reference `GPU_TARGETS` instead of `AMDGPU_TARGET`
* 68072698 fix: fix host generators not always supporting large sizes
* 50dab4f9 refactor: remove deprecated internal headers
* 80c16d40 test: improve accuracy of poisson histogram test
* 1a44b48c docs: fix styling

This PR does not contain fixes for the recent performance regressions, we'll either add them here or create a new PR for those.

PR #541, opened 2024-08-26
prefer the use of cdf over alias on gfx10xx and gfx11xx

Naraenda:
This MR resolves a performance regression on RDNA2 and RDNA3 platforms introduced with https://github.com/ROCm/rocRAND/pull/512 by reverting that change specifically on the aforementioned platforms.


NguyenNhuDi (approved):
lgtm

stanleytsang-amd (approved):

PR #557, opened 2024-09-26
Fixed memory leaks in rocrand_tests

NguyenNhuDi:
Added missing hipFree calls to test_rocrand_config_dispatch and test_rocrand_cpp_basic. Fixed an issue where test_utils was calling an extra hipCreateGraph call  causing a graph to be overwritten which led to a memory leak.

### DIFF
--- a/test/test_rocrand_hipgraphs.cpp
+++ b/test/test_rocrand_hipgraphs.cpp
@@ -34,34 +34,33 @@ void test_float(std::function<rocrand_status(rocrand_generator, float*, size_t, 
     HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
     rocrand_set_stream(generator, stream);
 
-    hipGraphExec_t graph_instance;
-    hipGraph_t graph = test_utils::createGraphHelper(stream);
+    test_utils::GraphHelper gHelper;
 
     // Any sizes
     ROCRAND_CHECK(
         generate_fn(generator, data, 1, mean, stddev)
     );
 
-    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
-    test_utils::resetGraphHelper(graph, graph_instance, stream);
+    gHelper.createGraph(stream, true, true, true);
+    gHelper.resetGraphHelper(stream);
 
     // Any alignment
     ROCRAND_CHECK(
         generate_fn(generator, data+1, 2, mean, stddev)
     );
 
-    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
-    test_utils::resetGraphHelper(graph, graph_instance, stream);
+    gHelper.createGraph(stream, true, true, true);
+    gHelper.resetGraphHelper(stream);
 
     ROCRAND_CHECK(
         generate_fn(generator, data, size, mean, stddev)
     );
 
-    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
+    gHelper.createGraph(stream, true, true, true);
 
     HIP_CHECK(hipFree(data));
     ROCRAND_CHECK(rocrand_destroy_generator(generator));
-    test_utils::cleanupGraphHelper(graph, graph_instance);
+    gHelper.cleanupGraphHelper();
     HIP_CHECK(hipStreamDestroy(stream));
 }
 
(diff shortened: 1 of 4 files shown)


test/test_rocrand_hipgraphs.cpp
```diff
@@ -34,34 +34,33 @@ void test_float(std::function<rocrand_status(rocrand_generator, float*, size_t,
     HIP_CHECK(hipStreamCreateWithFlags(&stream, hipStreamNonBlocking));
     rocrand_set_stream(generator, stream);
 
-    hipGraphExec_t graph_instance;
-    hipGraph_t graph = test_utils::createGraphHelper(stream);
+    test_utils::GraphHelper gHelper;
```
umfranzw, line 37: In order to capture the generate_fn() below in the graph, it needs to be in-between a call to hipStreamBeginCapture() and a call to hipStreamEndCapture().
Right now it looks like we're missing a call to hipStreamBeginCapture() before generate_fn().

I'd suggest making another member function in test_utils::GraphHelper - maybe something called startCapture() that calls hipStreamBeginCapture(). Then you can call it right after instantiating gHelper here.

gHelper.createGraph() below calls hipEndStreamCapture(), so I think then we'd be set.

This same change would need to be made in the other tests as well.

test/test_rocrand_hipgraphs.cpp
```diff
@@ -34,34 +34,33 @@ void test_float(std::function<rocrand_status(rocrand_generator, float*, size_t,
...
 
     // Any sizes
     ROCRAND_CHECK(
         generate_fn(generator, data, 1, mean, stddev)
     );
 
-    graph_instance = test_utils::endCaptureGraphHelper(graph, stream, true, true);
-    test_utils::resetGraphHelper(graph, graph_instance, stream);
+    gHelper.createGraph(stream, true, true, true);
```
umfranzw: This is just a small thing, but I noticed that this function also launches the graph - could we change the name slightly to make that clear (eg. createAndLaunchGraph())?

test/test_utils_hipgraphs.hpp
```diff
@@ -28,64 +28,53 @@
 // Note: graphs will not work on the default stream.
 namespace test_utils
 {
+    class GraphHelper{
+        private:
+            hipGraph_t graph;
+            hipGraphExec_t graph_instance;
+        public:
+            void createGraph(hipStream_t & stream, const bool beginCapture = true, const bool launchGraph=false, const bool sync=false){
+                if (beginCapture)
```
umfranzw: I think the beginCapture parameter and this "if" statement are no longer needed, since this function now ends the capture instead of starting it.

test/test_utils_hipgraphs.hpp
```diff
@@ -28,64 +28,53 @@
...
+                HIP_CHECK_NON_VOID(hipGraphExecDestroy(this->graph_instance));
+            }
+
+            inline void resetGraphHelper(hipStream_t& stream, const bool beginCapture=true)
+            {
+                // Destroy the old graph and instance
+                cleanupGraphHelper();
+
+                // Create a new graph and optionally begin capture
+                createGraph(stream, beginCapture);
```
umfranzw: After we cleanup the old graph and instance (by calling cleanupGraphHelper above), we will need to restart the capture again. 

createGraph actually ends capture and launches the graph, so that's not really what we need here. If you make a new member function that calls hipStreamBeginCapture as per the comments above, I think you probably could call that here instead.

umfranzw (changes requested):
I took a look through the change - it's really nice to have the helper as a class. I just have a few suggestions for some more changes to the way a few of the new helper member functions work.

test/test_utils_hipgraphs.hpp
NguyenNhuDi: Hi Wayne thanks for pointing my mistake out. I have added a startStreamCapture  in resetGraphHelper

test/test_utils_hipgraphs.hpp
NguyenNhuDi: got rid of this if , and added a separate function to be called before

test/test_rocrand_hipgraphs.cpp
NguyenNhuDi: I have changed the name to createAndLaunchGraph

test/test_utils_hipgraphs.hpp
```diff
@@ -28,64 +28,61 @@
...
+
+            inline void resetGraphHelper(hipStream_t& stream, const bool beginCapture=true)
+            {
+                // Destroy the old graph and instance
+                cleanupGraphHelper();
+
+                if(beginCapture)
+                    startStreamCapture(stream);
+
+                createAndLaunchGraph(stream);
```
umfranzw: I believe this call will need to be removed - it ends stream capture, which we don't want to do in this function.

test/test_utils_hipgraphs.hpp
```diff
@@ -28,64 +28,61 @@
 // Note: graphs will not work on the default stream.
 namespace test_utils
 {
+    class GraphHelper{
+        private:
+            hipGraph_t graph;
+            hipGraphExec_t graph_instance;
+        public:
+
+            inline void startStreamCapture(hipStream_t & stream){
```
umfranzw: There are a few opening curly braces here and in the following functions that could probably be moved to the next line (just to maintain consistency with the rest of the file).

umfranzw (changes requested):
I think this looks good - I've just mentioned two more small things I noticed below.

stanleytsang-amd (approved):

umfranzw (approved):

PR #563, opened 2024-10-09
Convert change log to new format

amd-jnovotny:
Changed the change log to the new format along with minor copy edits.


amd-jnovotny:
No longer necessary.

PR #577, opened 2024-10-28
Added gfx1151 support for -a default value in rmake.py

NguyenNhuDi:


rmake.py
```diff
@@ -116,7 +119,10 @@ def config_cmd():
     tools = f"-DCMAKE_TOOLCHAIN_FILE={toolchain}"
     cmake_options.append( tools )
 
-    cmake_options.extend( cmake_platform_opts)
+    if os.name == 'nt':
```
umfranzw: It looks like the reason that cmake_platform_opts is a sometimes a string instead of a list is because of line 117:
`cmake_platform_opts = f"-DROCM_DIR:PATH={rocm_path} -DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}"`

Could we change that line to:
`cmake_platform_opts.append(f"-DROCM_DIR:PATH={rocm_path} -DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}")`

I think that would make sure that it's always a list, regardless of what platform we're on. Then we wouldn't need the new if/else block that was added here (we could just go back to the original code):
`cmake_options.extend(cmake_platform_opts)`
Feel free to let me know what you think.

umfranzw (changes requested):

rmake.py
NguyenNhuDi: I think making cmake_platform_opts in line 117 to be a list  like so:
`cmake_platform_opts = [f"-DROCM_DIR:PATH={rocm_path}", f"-DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}"]`
makes more sense then doing
`cmake_platform_opts.append(f"-DROCM_DIR:PATH={rocm_path} -DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}")`, since cmake_options would eventually have an element that holds two commands instead of one command per elemenet

umfranzw: Good point - I hadn't noticed there were two options being added at the same time there. I agree that keeping them split makes sense.

PR #601, opened 2025-01-06
Update license file for 2025

amd-jnovotny:
Updated the copyright date in LICENSE.txt. This is the only change.


umfranzw (approved):
Looks good to me. This license is also included at the top of most source files, so we'll probably eventually need to update the year there as well, but we can follow-up with a separate change for that.

PR #627, opened 2025-03-17
[rocRAND] fast forwarding staging to head of develop

NguyenNhuDi:
* Added ROCM-Examples
* Rearranged include ordering in Dependencies.cmake so that CMAKE_INSTALL_LIBDIR will be lib on RHEL8


stanleytsang-amd:
If we're just fast forwarding staging from develop then it doesn't require a PR.  Normally develop gets autopromoted to staging but I guess that's been disabled now that the release-staging branch has been created.  I'll FF it myself.

PR #628, opened 2025-03-17
[rocRAND] fast forwarding rocm-rel-6.5 to head of develop

NguyenNhuDi:
* Added ROCM-Examples
* Rearranged include ordering in Dependencies.cmake so that CMAKE_INSTALL_LIBDIR will be lib on RHEL8


stanleytsang-amd:
If we're fast-forwarding, in the absence of autopromotion this is something that I can do locally and push to the github branches.  Cleaner linear commit history between all branches.

PR #630, opened 2025-03-20
[rocRAND] Made default of install benchmark to be ON when BUILD_BENCHMARK=ON

NguyenNhuDi:


NguyenNhuDi:
@Naraenda was there a reason why the flag was set to be OFF by default?

Naraenda (approved):
Change is fine. NTA

Naraenda:
> @Naraenda was there a reason why the flag was set to be OFF by default?

Not that I'm aware of. I guess it's more proper to have things disabled by default, but I don't have strong opinions.

stanleytsang-amd (approved):

PR #635, opened 2025-04-11
Move restore user global state code to proper section

stanleytsang-amd:
I believe #624 was incomplete in that it did not move the restore user global state code https://github.com/ROCm/rocRAND/blob/9203a70993c3dc876eb0d45055082ab12fa58e39/cmake/Dependencies.cmake#L160
to after the configuration of googletest and googlebenchmark.

This results in the tests and benchmarks to be built as shared libraries by default and not static libraries (which is how it worked before).  This then requires googletest/bench libs to be installed on the system, which is not guaranteed and is causing issues in QA testing. 

@NguyenNhuDi Please confirm if the location of the restore user global state code was intended by your PR.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -103,15 +103,6 @@ if(NOT ROCM_FOUND)
   find_package(ROCM 0.7.3 REQUIRED CONFIG PATHS ${PROJECT_EXTERN_DIR} NO_DEFAULT_PATH)
 endif()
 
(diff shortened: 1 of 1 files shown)


RobsonRLemos (approved):
Looks good to me. The failing check seems not to be related to the modifications.

NguyenNhuDi (approved):
LGTM. The location of restore user global state was missed in my last PR (#624)

